Issue #122, opened 2015-01-02
Installer CLI not searching for plugins correctly


sloria:
When I run `mfr install all`, it appears that the CLI is not searching for plugins correctly. Here's what I get when I have my osf virtualenv enabled. 

```
$ mfr install all
Plugin with name "_billiard.so" not found. Skipping...
Plugin with name "_regex.so" not found. Skipping...
Plugin with name "_regex_core.py" not found. Skipping...
Plugin with name "_regex_core.pyc" not found. Skipping...
Plugin with name "argparse.py" not found. Skipping...
Plugin with name "argparse.pyc" not found. Skipping...
Plugin with name "backports.ssl_match_hostname-3.4.0.2-py2.7.egg-info" not found. Skipping...
Plugin with name "cssmin.py" not found. Skipping...
Plugin with name "cssmin.pyc" not found. Skipping...
Plugin with name "dataverse.egg-link" not found. Skipping...
Plugin with name "decorator.py" not found. Skipping...
Plugin with name "decorator.pyc" not found. Skipping...
Plugin with name "docopt.py" not found. Skipping...
Plugin with name "docopt.pyc" not found. Skipping...
Plugin with name "easy-install.pth" not found. Skipping...
Plugin with name "easy_install.py" not found. Skipping...
Plugin with name "easy_install.pyc" not found. Skipping...
Plugin with name "gig.py" not found. Skipping...
Plugin with name "gig.pyc" not found. Skipping...
Plugin with name "gnupg.py" not found. Skipping...
Plugin with name "gnupg.pyc" not found. Skipping...
Plugin with name "gnureadline.so" not found. Skipping...
Plugin with name "hurry.filesize-0.9-py2.7-nspkg.pth" not found. Skipping...
Plugin with name "ipython-2.1.0-py2.7.egg-info" not found. Skipping...
Plugin with name "itsdangerous.py" not found. Skipping...
Plugin with name "itsdangerous.pyc" not found. Skipping...
Plugin with name "magic.py" not found. Skipping...
Plugin with name "magic.pyc" not found. Skipping...
Plugin with name "mailchimp.py" not found. Skipping...
Plugin with name "mailchimp.pyc" not found. Skipping...
Plugin with name "mccabe.py" not found. Skipping...
Plugin with name "mccabe.pyc" not found. Skipping...
[...]

PR #123, opened 2015-01-02
Make CLI install modules matching 'mfr_[module]'

mfraezz:
Purpose
=======
* Limit dirs/files searched to only directories with valid naming convention when calling <code>mfr install all<code>
* Closes https://github.com/CenterForOpenScience/modular-file-renderer/issues/122

Changes
=======
Ignores non-directories and non-matching directories when installing plugins

Side Effects
==========
Removes unnecessary render-requirements.txt from mfr_pdb

### DIFF
--- a/mfr/cli.py
+++ b/mfr/cli.py
@@ -2,6 +2,7 @@
 """Command Line tool for the modular file renderer"""
 
 import os
+import re
 import pip
 import sys
 import argparse
@@ -32,7 +33,10 @@ def plugin_requirements(render, export, plugins):
 
     if plugins == ["all"]:
         path_list = [os.path.join(ROOT_PATH, directory)
-                     for directory in os.listdir(ROOT_PATH)]
+                     for directory in os.listdir(ROOT_PATH)
+                        if bool(re.match('mfr_\w*',
+                            directory.split('/')[len(directory.split('/')) - 1]))]
+
     else:
         path_list = [os.path.join(ROOT_PATH, plugin) for plugin in plugins]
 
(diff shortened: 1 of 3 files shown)


mfr/cli.py
```diff
@@ -32,7 +33,10 @@ def plugin_requirements(render, export, plugins):
 
     if plugins == ["all"]:
         path_list = [os.path.join(ROOT_PATH, directory)
-                     for directory in os.listdir(ROOT_PATH)]
+                     for directory in os.listdir(ROOT_PATH)
+                        if bool(re.match('mfr_\w*',
+                            directory.split('/')[len(directory.split('/')) - 1]))]
```
sloria: I believe you can use `os.path.basename` here.

mfr_pdb/render.py
```diff
@@ -1,11 +1,12 @@
 """Molecule renderer module """
-
+#Uses PV: https://github.com/biasmv/pv
 import os
 import mfr
 from mfr.core import RenderResult, get_assets_from_list
 
 # assets must be loaded in this order
 JS_ASSETS = [
+    #The OSF has jquery on each page, uncomment if necessary (e.g. for the player)
```
sloria: The user should not have to modify any source files manually. There should either be a config option to include jquery, or add a note in the documentation that jquery is required for this renderer.

mfr/cli.py
mfraezz: You're completely right. Fixing this now

mfr_pdb/render.py
mfraezz: Got it. I'll add a note in the documentation.

PR #128, opened 2015-01-22
Hotfix/tci builds

mfraezz:
Purpose
======
Fix build errors caused by recent refactoring.

Changes
=======
* Discontinue use of <code>csv</code> library.
* Use <code>pandas</code> for .tsv rendering
* Add test, test file for .tsv rendering
* Update image test path to new naming convention

Side Effects
=========
* .tsv rendering looks better than it did


mfr/ext/image/tests/test_image.py
```diff
@@ -39,11 +39,11 @@ def test_render_img_tag(fakefile):
     assert 'alt="My image"' in result
 
 
-@pytest.mark.skipif(sys.version_info > (2, 7), reason="requires python2.7 or less")
+@pytest.mark.skipif(sys.version_info > (2, 8), reason="requires python2.7 or less")
```
sloria: Why was this changed?

mfraezz: It was being skipped in 2.7 tests, but the exporter should work in 2.6 and 2.7.

PR #129, opened 2015-01-23
Fix hard-coded static path issue

mfraezz:
Purpose
========
Having hard-coded absolute paths is poor practice, resolving them dynamically is less likely to fail if anything is changed

Changes
=======
* Make <code>collect_static()</code> resolve dependencies when collecting
* Template once hard-coded static files

Side Effects
==========
 <code>collect_static()</code> takes slightly longer to run


mfr/core.py
```diff
@@ -370,3 +371,26 @@ def collect_static(dest=None, dry_run=False):
...
+            template_path = os.path.join(namespaced_destination, 'templates/')
+            # If there are templated static files that once had hard-coded dependencies
+            if os.path.isdir(template_path):
+                for dirpath, dirnames, filenames in os.walk(template_path):
+                    # Walk through each template
+                    for template in filenames:
+                        with open(os.path.join(template_path, template)) as src:
+                            # Resolve the dependency
+                            data = re.sub('{{STATIC_PATH}}',
+                                '/static'+namespaced_destination.split('static')[-1],
```
sloria: This will not work if the user configures mfr.config['STATIC_PATH'] to anything other than `/static`. Why not replace `{{STATIC_PATH}}` with the config variable?

sloria:
There is a lot of code for templating the static assets here; I'm wondering if it can be made simpler *and* more general; for example, why only allow `{{STATIC_PATH}}`? Why not pass the whole mfr config object (could use a templating language like Jinja or Mako), e.g. `{{ mfr.config.STATIC_PATH}}`?

Holding off on merging this for now. Will just fix the hardcoded paths until we have more time to implement a better solution for templating assets.

PR #130, opened 2015-01-27
Dont load older versions of jquery as it mucks things

chrisseto:


sloria:
I've done this in https://github.com/CenterForOpenScience/modular-file-renderer/pull/132 . Thanks though

PR #137, opened 2015-01-27
Hotfix/xlsx issue

mfraezz:


mfr/ext/tabular/libs/xlrd_tools.py
```diff
@@ -26,7 +26,11 @@ def xlsx_xlrd(fp):
 
     fields = sheet.row_values(0) if sheet.nrows else []
 
-    fields = [str(value.encode('ascii', 'ignore')) or 'Unnamed: {0}'.format(index+1) for index, value in enumerate(fields)]
+    fields = [
+        str(value)
```
sloria: What if value is `None`?

sloria: I think the logic for unnamed fiields is incorrect. If `value` is a string, then you don't need to default it to `Unnamed: {idx}`.

sloria: Perhaps separate this field "cleaning" into a separate function so that you can unit test this.

Issue #138, opened 2015-01-28
collect_static allows OSErrors to pass silently


sloria:
`collect_static` is supposed to skip folders that have already copied. That happens here:

https://github.com/CenterForOpenScience/modular-file-renderer/blob/113d71ad9a8b7a5297b3946b6054639add2cf76b/mfr/core.py#L340

an unwanted consequence of this is that unrelated OSErrors (such as permissions errors) are allowed to pass silently.

We should *only* pass if the copy destination directory already exists, not for any other reason.

PR #139, opened 2015-01-28
Hotfix/static collecting

mfraezz:
Purpose
=======
OSErrors unrelated to files already existing (such as permissions errors) caught during collect_static() should not be allowed to pass silently.
Closes https://github.com/CenterForOpenScience/modular-file-renderer/issues/138

Changes
========
When an OSError is caught, check to see if the path exists. If not, it's an unrelated error and is raised.

Side Effects
==========
OSErrors might be raised.

### DIFF
--- a/mfr/core.py
+++ b/mfr/core.py
@@ -338,8 +338,8 @@ def copy_dir(src, dest):
     except shutil.Error as err:
         logger.warn(err)
     except OSError as err:
-        if os.path.exists(src):
-            logger.debug('Skipping {src} (already exists)'.format(src=src))
+        if os.path.exists(dest):
+            logger.debug('Skipping {dest} (already exists)'.format(dest=dest))
         else:
             raise err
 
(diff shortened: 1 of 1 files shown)


mfr/core.py
```diff
@@ -338,7 +338,10 @@ def copy_dir(src, dest):
     except shutil.Error as err:
         logger.warn(err)
     except OSError as err:
-        logger.debug('Skipping {src} (already exists)'.format(src=src))
+        if os.path.exists(src):
```
sloria: This will always be `True`. The `src` file will always exist. You should check the existence of `dest`.

sloria: Instead of using `os.path.exists` (which is prone to [race conditions](https://stackoverflow.com/questions/14574518/how-does-using-the-try-statement-avoid-a-race-condition), can you inspect some property of `err`?

sloria:
Pass finished. :tractor:

mfraezz:
Pass comment fixed

PR #143, opened 2015-02-11
[feature] wheelhouse environment

icereval:
Purpose
---

Speed up travis-ci builds in OSF repo.

Changes
---

Added WHEELHOUSE environment variable.
Modified install script to reference the wheel path.

Side effects
---

If travis-ci changes their server builds or we update requirements, we'll need to build the corresponding whl files and update our repository.

Worst case scenario the build time should simply increase until we update the wheelhouse repo.

### DIFF
--- a/mfr/cli.py
+++ b/mfr/cli.py
@@ -10,7 +10,7 @@ HERE = os.path.dirname(os.path.abspath(__file__))
 EXT_PATH = os.path.join(HERE, 'ext')
 
 
-def pip_install(path, filename):
+def pip_install(path, filename, wheelhouse=None):
     """Use pip to install from a requirements file
 
     :param path: location of file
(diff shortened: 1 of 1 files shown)


icereval:
##### Times from the following commands:

`travis_retry invoke requirements`
`travis_retry invoke addon_requirements`
`travis_retry mfr -r install all`

*We can likely add additional dependancies to our wheelhouse to speed these up further*

Before
----

image


After
---

image

mfr/cli.py
```diff
@@ -19,7 +20,10 @@ def pip_install(path, filename):
     file_location = (os.path.join(path, filename))
 
     if os.path.isfile(file_location):
-        pip.main(['install', "-r", file_location])
+        pip_args = ['install', '-r', file_location]
+        if WHEELHOUSE_PATH:
```
sloria: We should add a  `--wheelhouse` or `--no-wheelhouse` option to override this behavior. 

This doesn't have to get in before this is merged.

PR #147, opened 2015-03-04
Handle unicode characters in table files; add regression test.

jmcarp:
The `strip_comments` helper was previously attempting to encode
(potentially) bytestrings to bytestrings, which caused rendering with
non-ascii characters to fail. This patch only decodes to bytes if
necessary, and updates the csv test fixture with a unicode character to
prevent future regressions.

@sloria @mfraezz: This fixes the problem I was running into on OSF, but open to a better solution if this one is not rad.


mfraezz:
Has a sufficient level of radness for my approval :tea:

PR #148, opened 2015-03-05
Fix installation

mfraezz:
Purpose
=======
Fix build when running `python setup.py install` or `pip install .` , and on TCI

Changes
=======
Allow for arbitrary depths of static assets

Side Effects
=========
None

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@
 
 language: python
 
+sudo: false
+
 python:
   - "2.6"
   - "2.7"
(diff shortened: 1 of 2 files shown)


sloria:
@mfraezz Can you `pip install .` to `.travis.yml` so that the installation is tested?

sloria:
By the way, you should also be able to add `sudo: false` to `.travis.yml` to use TCI's new container-based system: http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

PR #154, opened 2015-03-09
[feature] tabular error handling

mfraezz:
Purpose
=======
Better error handling when rendering tabular files

Changes
=======
* new exception class
* try/excepts at points where errors arising from non-ascii characters can occur

Side Effects
==========
None

### DIFF
--- a/mfr/ext/tabular/render.py
+++ b/mfr/ext/tabular/render.py
@@ -55,8 +55,8 @@ def render_html(fp, src=None):
     slick_grid_options = config.get('slick_grid_options').get(table_size)
 
     try:
-        columns = json.dumps(columns),
-        rows = json.dumps(rows),
+        columns = json.dumps(columns)
+        rows = json.dumps(rows)
     except UnicodeDecodeError:
         raise UnexpectedFormattingException()
 
--- a/mfr/ext/tabular/render.py
+++ b/mfr/ext/tabular/render.py
@@ -9,7 +9,8 @@ import mfr
 import os
 from .configuration import config
 from .exceptions import TableTooBigException, \
-    EmptyTableException, MissingRequirementsException
+    EmptyTableException, MissingRequirementsException, \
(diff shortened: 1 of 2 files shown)


mfr/ext/tabular/render.py
```diff
@@ -50,12 +54,18 @@ def render_html(fp, src=None):
     table_size = 'small_table' if len(columns) < 9 else 'big_table'
     slick_grid_options = config.get('slick_grid_options').get(table_size)
 
+    try:
+        columns = json.dumps(columns)
+        rows = json.dumps(rows)
+    except UnicodeDecodeError:
+        raise UnexpectedFormattingException()
```
sloria: Add a descriptive error message.

mfraezz: The default error message that will be displayed is `Unable to render. Download(link) file to view it. [any additional msg goes here]`

mfr/ext/tabular/render.py
```diff
@@ -35,7 +36,10 @@ def render_html(fp, src=None):
     :return: RenderResult object containing html and assets
     """
 
-    columns, rows = populate_data(fp)
+    try:
+        columns, rows = populate_data(fp)
+    except KeyError:
```
sloria: I'm thinking this `KeyError` should be handled lower in the call stack; maybe in one of the function in `libs`, or--if that produced lots of duplication--in `populate_data`.

sloria:
Pass finished. :ocean:

mfr/ext/tabular/render.py
mfraezz: Good call. The key in question tends to be a hex value (e.g. `KeyError: '\x94'`), and while in testing it I've only found KeyErrors raised during the rendering of .dta files, I don't know for sure that it's limited to them. `populate_data` seems the safer place to put it if we want it to cover all tabular filetypes without all the duplication.

PR #156, opened 2015-03-10
[hold] Feature/patch pydocx parsed

mfraezz:
Purpose
=======
Override the functions that provide breaking html from external library PyDocX

Changes
=======
* Class added
* Dependency changed

Side Effects
=========
None


mfraezz:
Wait on merging this PR, the people maintaining PyDocX may decide to add a property that would obviate it.

PR #157, opened 2015-03-12
Feature/fix packagedata

mfraezz:
Purpose
=======
Allow `python setup.py install` to collect all the proper static dependencies

Changes
=======
Be more explicit when defining setup()'s package_data param

Side Effect
========
None


setup.py
```diff
@@ -66,7 +67,16 @@ def read(fname):
     ),
     include_package_data=True,
     package_data={
-        "": ['templates/*', 'static/*./*', '*.js', '*.css', '*.txt'],
+        "mfr.ext.code_pygments": ['static/css/*'],
+        "mfr.ext.ipynb": ['static/*/*', 'templates/*'],
+        "mfr.ext.pdb": ['static/*/*', 'templates/*']
+        "mfr.ext.pdf": ['static/build/*',
+                        'static/web/*.*'
```
sloria: Why the `.`?

sloria: Also, missing a comma here.

mfraezz: There are several files of different extensions in `web`, and doing `static/web/*` would attempt to copy the directories therein as well.

PR #158, opened 2015-03-12
Fix package_data static collection

mfraezz:
See https://github.com/CenterForOpenScience/modular-file-renderer/pull/157


setup.py
```diff
@@ -66,7 +66,16 @@ def read(fname):
     ),
     include_package_data=True,
     package_data={
-        "": ['templates/*', 'static/*./*', '*.js', '*.css', '*.txt'],
+        "mfr.ext.code_pygments": ['static/css/*'],
+        "mfr.ext.ipynb": ['static/*/*', 'templates/*'],
+        "mfr.ext.pdb": ['static/*/*', 'templates/*'],
+        "mfr.ext.pdf": ['static/build/*',
+                        'static/web/*.*',
```
sloria: These paths aren't correct; the new PDF renderer is not on master

PR #167, opened 2015-04-03
Decode unicode chars in pygments files iff necessary

mfraezz:
Purpose
=======
Closes https://github.com/CenterForOpenScience/osf.io/issues/2389

Changes
=======
* .is_printable() added to check if decoding is necessary
* decodes iff necessary

Side Effects
=========
None


sloria:
MFR should not make any assumptions about the encoding of file content; that is the responsibility of the user of the MFR. Hence, https://github.com/CenterForOpenScience/osf.io/commit/8cd0b05f87c9ce32139196262a95ddae593f029a.

PR #178, opened 2015-07-28
Detect quotechars and doublequotes in csv dialect

mfraezz:
Purpose
=======
Closes https://github.com/CenterForOpenScience/osf.io/issues/3817

Changes
=======
Add helper method to augment stdlib's csv.Sniffer for special cases.
Uses regex to look for comma-separated lists inside csv's

Side Effects
==========
None

### DIFF
--- a/mfr/extensions/tabular/libs/stdlib_tools.py
+++ b/mfr/extensions/tabular/libs/stdlib_tools.py
@@ -1,4 +1,5 @@
 import csv
+import re
 
 
 def csv_stdlib(fp):
@@ -37,13 +38,17 @@ def csv_stdlib(fp):
 
 def detect_quoted_fields(dialect, data):
     if dialect != csv.excel:
-        if data.find("''") >= 0:
-            dialect.doublequote = True
-            dialect.quotechar = "'"
+        if dialect.quotechar == '"':
+            if re.search('\'[[({]\".+\",', data):
+                dialect.quotechar = "'"
+            if re.search('\'\'\'[[({]\".+\",', data):
+                dialect.doublequote = True
             return dialect
-        elif data.find('""') >= 0:
-            dialect.doublequote = True
-            dialect.quotechar = '"'
+        elif dialect.quotechar == "'":
+            if re.search('\"[[({]\'.+\',', data):
+                dialect.quotechar = '"'
+            if re.search('\"\"\"[[({]\'.+\',', data):
+                dialect.doublequote = True
             return dialect
 
     return dialect
--- a/mfr/extensions/tabular/libs/stdlib_tools.py
+++ b/mfr/extensions/tabular/libs/stdlib_tools.py
@@ -6,12 +6,15 @@ def csv_stdlib(fp):
     :param fp: File pointer object
     :return: tuple of table headers and data
     """
+    data = fp.read(2048)
+    fp.seek(0)
+
     try:
-        dialect = csv.Sniffer().sniff(fp.read(2048))
+        dialect = csv.Sniffer().sniff(data)
     except csv.Error:
         dialect = csv.excel
 
-    fp.seek(0)
+    dialect = detect_quoted_fields(dialect, data)
     reader = csv.DictReader(fp, dialect=dialect)
     columns = []
     # update the reader field names to avoid duplicate column names when performing row extraction
@@ -30,3 +33,17 @@ def csv_stdlib(fp):
         })
     rows = [row for row in reader]
     return columns, rows
+
+
+def detect_quoted_fields(dialect, data):
+    if dialect != csv.excel:
+        if data.find("''") >= 0:
+            dialect.doublequote = True
+            dialect.quotechar = "'"
+            return dialect
+        elif data.find('""') >= 0:
+            dialect.doublequote = True
+            dialect.quotechar = '"'
+            return dialect
+
+    return dialect
--- a/mfr/extensions/tabular/libs/stdlib_tools.py
+++ b/mfr/extensions/tabular/libs/stdlib_tools.py
@@ -49,12 +49,12 @@ def _set_dialect_quote_attrs(dialect, data):
         '{"a", "b", "c"         for quotechar == "
     """
     if dialect.quotechar == '"':
-        if re.search('\'[[({]\".+\",', data):
+        if re.search('\'[[({]".+",', data):
             dialect.quotechar = "'"
-        if re.search('\'\'\'[[({]\".+\",', data):
+        if re.search("'''[[({]\".+\",", data):
             dialect.doublequote = True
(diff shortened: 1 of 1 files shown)


lbanner:
Ping @lyndsysimon -- Please prioritize this one. @sloria This is a fix we need to get out as soon a possible in support of a user's efforts. Thanks.

mfr/extensions/tabular/libs/stdlib_tools.py
```diff
@@ -30,3 +34,21 @@ def csv_stdlib(fp):
         })
     rows = [row for row in reader]
     return columns, rows
+
+
+def detect_quoted_fields(dialect, data):
```
sloria: Is this necessary, or can we take advantage of the quoting options in the `csv` module? https://docs.python.org/2/library/csv.html#csv.QUOTE_ALL

sloria: If this function *is* necessary, its name should indicate that it modifies `dialect`; "detect" implies a read-only action. Perhaps `set_dialect_quotechars`.

sloria: Also, the regexes should be compiled for performance, using `re.compile`.

But again, it would be more ideal if we didn't have to do regex-matching at all.

mfr/extensions/tabular/libs/stdlib_tools.py
```diff
@@ -6,12 +7,15 @@ def csv_stdlib(fp):
...
+    fp.seek(0)
+
     try:
-        dialect = csv.Sniffer().sniff(fp.read(2048))
+        dialect = csv.Sniffer().sniff(data)
     except csv.Error:
         dialect = csv.excel
 
-    fp.seek(0)
+    dialect = detect_quoted_fields(dialect, data)
```
lyndsysimon: Let's move this line to an `else` block on the `try...except`, and remove the conditional in `detect_quoted_fields`.

mfr/extensions/tabular/libs/stdlib_tools.py
sloria: What are the default quoting characters? Is it possible that the user needs to escape the quotes a certain way?

mfr/extensions/tabular/libs/stdlib_tools.py
```diff
@@ -30,3 +34,21 @@ def csv_stdlib(fp):
         })
     rows = [row for row in reader]
     return columns, rows
+
+
+def detect_quoted_fields(dialect, data):
+    if dialect != csv.excel:
+        if dialect.quotechar == '"':
+            if re.search('\'[[({]\".+\",', data):
```
lyndsysimon: Please add comments to explain the purpose of these regexes

mfr/extensions/tabular/libs/stdlib_tools.py
```diff
@@ -30,3 +34,21 @@ def csv_stdlib(fp):
...
+                dialect.quotechar = "'"
+            if re.search('\'\'\'[[({]\".+\",', data):
+                dialect.doublequote = True
+            return dialect
```
lyndsysimon: These return statements are superfluous, with a return at the end of the function.

mfr/extensions/tabular/libs/stdlib_tools.py
lyndsysimon: If this function remains, please add a docstring.

lyndsysimon:
I'd normally ask that changes like this be accompanied by a regression test - but in this case, I don't see that the `csv_stdlib` method is tested at all. That being the case, I'm hesitant to require it.

That said, writing a unit test for `detect_quoted_fields` would be a reasonable incremental approach.

I'm not going to make this a blocker for me, so... if it's more overhead than it's worth to test it, don't. It wil be caught when a test is written for the `csv_stdlib` function.

mfr/extensions/tabular/libs/stdlib_tools.py
mfraezz: Cannot take advantage of quoting options. Default is `QUOTE_MINIMAL`, which should be (and is when `dialect.quotechar` is set properly) handling the management of delimiters inside of fields. 

The issue is that the sniffer detects the `dialect.quotechar` incorrectly for large lists of non-quotechar-wrapped items inside of the actual quotechar. I tried numerous test files, including ones in which all quotation marks, single or double, were escaped. 

Regex use is necessary to catch all possible formatting versions of this type of bug.

mfraezz: Regarding `re.compile`'ing, that would only yield a performance benefit if it was done in the global scope. Compiling four regexes each time this function is called would, at best, yield no change in performance. I'd also argue that this function isn't called frequently enough to merit global regex compiling

mfraezz: Quoting characters are determined by the sniffer, seemingly based on frequency of occurrence (either `'` or `"`).

mfr/extensions/tabular/libs/stdlib_tools.py
```diff
@@ -30,3 +35,26 @@ def csv_stdlib(fp):
         })
     rows = [row for row in reader]
     return columns, rows
+
+
+def set_dialect_quote_attrs(dialect, data):
```
icereval: private methods should be prefixed with `_`

mfr/extensions/tabular/libs/stdlib_tools.py
```diff
@@ -30,3 +35,26 @@ def csv_stdlib(fp):
...
+    list items wrapped in dialect.quotechar and seperated by commas.
+
+    Example matches include:
+        "['1', '2', '3'         for quotechar == '
+        '{"a", "b", "c"         for quotechar == "
+    """
+    if dialect.quotechar == '"':
+        if re.search('\'[[({]\".+\",', data):
+            dialect.quotechar = "'"
+        if re.search('\'\'\'[[({]\".+\",', data):
```
chrisseto: When writing regexes avoid escaping characters as much as possible.
Use raw strings, `r'...'` and swap double and single quotes as need be.
`r"'''[[({]\".+\","`

mfraezz: Fixed. 

Why using `r'...'` is a bad idea: [last paragraph](https://docs.python.org/2.0/ref/strings.html)

I could also check for the termination of the list, but I didn't because there's always a chance that it would exceed the 2kb read in to `data`.

icereval:
This has been released, merged in [hotfix 0.3.7](https://github.com/CenterForOpenScience/modular-file-renderer/commit/f43c6dffa108b01e83ec180df77310686cf2094e)

PR #179, opened 2015-08-06
[Feature Fix] Refine the warning message about not supportting rendering

haoyuchen1992:
### Purpose & Change
Change the warning message when dealing with non-supported file type
Old:
`No renderer could be found for the file type requested.`
New:
`Viewing of this file type within the OSF is not currently supported. Please download the file to view.`

Resolve https://github.com/CenterForOpenScience/osf.io/issues/3907
### Side Effect
None

### Other related issues
The similar vague error [message](https://github.com/CenterForOpenScience/modular-file-renderer/blob/bb325ec753045dddd587abc4b23747ccb21a4d02/mfr/core/utils.py#L69): `No exporter could be found for the file type requested`. It may also need change.

### DIFF
--- a/mfr/core/utils.py
+++ b/mfr/core/utils.py
@@ -89,4 +89,4 @@ def make_renderer(name, metadata, file_path, url, assets_url, export_url):
             invoke_args=(metadata, file_path, url, assets_url, export_url),
         ).driver
     except RuntimeError:
-        raise exceptions.RendererError('No renderer could be found for the file type requested.', code=400)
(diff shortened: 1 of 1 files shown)


haoyuchen1992:
@icereval Here is a simple PR  and ready to be merged!

sloria:
There should be no mention of the "OSF" or any OSF-related code in MFR; it is a standalone service.

Instead, OSF-specific messages should be handled in the OSF. Please revise the approach.

PR #180, opened 2015-08-07
Add rpy2 to render .sav files

mfraezz:
Purpose
=======
Closes https://github.com/CenterForOpenScience/osf.io/issues/2149
Allows for rendering of .sav (SPSS data) files. The code to do so is already present, but the library `rpy2`, required by pandas, is not being installed.

Changes
=======
Add `rpy2` to `requirements.txt`

Current .sav file behavior:
-------------------------------------
screen shot 2015-08-07 at 11 55 57

New behavior:
-------------------
screen shot 2015-08-07 at 11 56 39

Side effects:
==========
`gcc` must be installed. It is on our mfr server, according to @mattclark, and it is by default on most linux distributions. Brew installable on OS X.
`R` must be installed. This poses an issue for TCI.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -32,4 +32,5 @@ docutils==0.12
 
 # Tabular
 pandas==0.16.1
+rpy2==2.6.1
 git+https://github.com/icereval/xlrd.git


mfraezz:
There appears to be no good way to install R for TCI. Will add note in a `README` somewhere about this.

PR #181, opened 2015-08-10
[Feature Fix] Refine MRF warning message not supporting rendering and exporting

haoyuchen1992:
### Purpose
The warning messages in MFR are too abstract to users. 
Resolve https://github.com/CenterForOpenScience/osf.io/issues/3907

### Changes
Move warning message into setting page and also change the text.

Change the warning message when dealing with non-supported file type
Old: `No renderer could be found for the file type requested.`
New: `Viewing of this file type is not currently supported. Please download the file to view.`

Change the warning message when dealing with non-exported file type
Old: `No exporter could be found for the file type requested.`
New: `Exporting of this file type is not currently supported.`

### Side Effect
None

### DIFF
--- a/mfr/core/utils.py
+++ b/mfr/core/utils.py
@@ -66,7 +66,7 @@ def make_exporter(name, source_file_path, output_file_path, format):
             invoke_args=(source_file_path, output_file_path, format),
         ).driver
     except RuntimeError:
-        raise exceptions.RendererError('No exporter could be found for the file type requested.', code=400)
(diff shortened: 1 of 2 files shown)


lbanner:
@lyndsysimon @sloria -- This should be a quick review -- Can we get  this fix in? Thanks.

mfraezz:
No mention of other services, and the messages are in a more sensible location. Looks good :star2:

Issue #177, opened 2015-07-20
Don't include Bootstrap in codepygments module


sloria:
I'm not sure that it's necessary to include Bootstrap for the codepygments module. We might just remove it.

Related issue: https://github.com/CenterForOpenScience/modular-file-renderer/issues/176 - If we do need to include Boostrap, it would still be good to have a way to opt out of it.

chrisseto:
+1
it looks like bootstrap is only used for the error messages.
Foundation is included and I think it could be removed as well.

mfraezz:
While it would prevent #175, some of the Boostrap rules are necessary. Otherwise, it will look like this:
![screen shot 2015-08-19 at 14 21 05](https://cloud.githubusercontent.com/assets/5659262/9365090/9f6ca1b6-467d-11e5-83fd-442cd0ba6dcc.png)

Misalignment and lack of a border are obvious, but it's also not horizontally scrollable. A simple solution would be to just include the 10 or so rules that pygments actually needs to render things nicely in `default.css`, and remove the Bootstrap dependency.

Agree with @chrisseto about Foundation, it doesn't appear to be used anywhere at all.

mfraezz:
In fact, nothing in `/pdb/css/` appears to be used.

Issue #176, opened 2015-07-20
Don't include vendor assets


sloria:
Some renderers load common vendor assets, e.g. codepygments includes Bootstrap. Often, render results will be included on a page that already includes these assets. It would be good to prevent these duplicate loads by allowing the user to opt in or opt out of including specific assets.

Issue #175, opened 2015-07-09
[Style Guideline] Need override the boarder with radius for pre tag


haoyuchen1992:
In style guideline, we only allow the border without radius. The pre tag in MFR create border with radius in osf. 

Here is [Trello bug](https://trello.com/c/kh9ClUOc/187-mfr-reader-file-render-border-has-rounded-corners-instead-of-square)

Example:
![radius](https://cloud.githubusercontent.com/assets/5420789/8606609/be338684-265a-11e5-897e-cc57db5e739b.gif)

PR #184, opened 2015-08-20
[Feature] [Fix] Cleaning static

mfraezz:
Purpose
=======
Closes https://github.com/CenterForOpenScience/modular-file-renderer/issues/177
Closes https://github.com/CenterForOpenScience/modular-file-renderer/issues/176
Closes https://github.com/CenterForOpenScience/modular-file-renderer/issues/175

Changes
=======
* Put bootstrap css rules used by `codepygments` in `default.css`
* Removed unnecessary static files in `/pdb/`, `/pdf/`, `/svg/`, and `/tabular/`
* Remove reference to a deleted example file
    * It would never be used, MFR wouldn't get that far without a file being passed in:
```javascript
var DEFAULT_URL = '[some research paper by Mozilla].pdf';
[...]
var file = 'file' in params ? params.file : DEFAULT_URL;
```

Side Effects
=========
None, tested all filetypes that could be affected

CR Note
------------
[Commit 2]( https://github.com/mfraezz/modular-file-renderer/commit/d0ab31e63ca21abc4dbacae963d48f8bc9fcd85a ) is exclusively deletions, view other commits for modifications


lyndsysimon:
Looks great. I've got some follow-up questions for you Steve, but they shouldn't be a part of this PR.

lyndsysimon:
Oh! I almost forgot the gold star: :star2:

icereval:
I'm struggling to understand why we would want to customize third party static assets? Third party plugins are meant to be drop in upgrades to the latest version with the fewest number of changes possible to have them integrated w/ the MFR.

This seems like a lot of work to maintain this practice going forward and also adds significant risk of an asset being missed or mistakenly removed.

If we are concerned about bloat in the code base we might investigate fixing this issue with a dependency manager like Bower, thus third party libraries are only brought into the picture on build and can remain separate.

Reviewing the referenced tickets they can be addressed with minimal changes such as removing the references to dependencies.

Please resubmit w/o asset deletions.

mfraezz:
I'll close this for now, get a PR up for the pygments issues and add an issue about using a dependency manager.

PR #186, opened 2015-08-24
[Feature][Fix][Ready For Review] Remove bootstrap dependency from Pygments

### LINKED ISSUE #175: [Style Guideline] Need override the boarder with radius for pre tag

### LINKED ISSUE #177: Don't include Bootstrap in codepygments module

mfraezz:
Purpose
=======
Closes #177 
Closes #175 

Changes
=======
* Bootstrap dependency removed
* Necessary CSS rules from bootstrap added

Side Effects
=========
None


lyndsysimon:
I see no bugs here. :star:

sloria:
Looks good to me.

PR #187, opened 2015-08-25
[Feature][Ready for Review] Tabular Filtering and Multisheet

### LINKED ISSUE #4: Render tabular data
Render csv, xls(x), r, sps, etc.

mfraezz:
Purpose
======
Closes https://github.com/CenterForOpenScience/osf.io/issues/500
Closes https://github.com/CenterForOpenScience/osf.io/issues/4102
Closes #4 

Changes
=======
* Support added for displaying multiple spreadsheets in a single file 
   * Scrolling tabs allow you to switch between sheets
* Support added for searching/filtering rows for certain string

Side Effects
=========
Filetypes that don't support multiple spreadsheets (e.g. csv, tsv) will have `Sheet_1` displayed in a single tab.


mfr/extensions/tabular/libs/ezodf_tools.py
```diff
@@ -12,11 +12,13 @@ def ods_ezodf(fp):
...
 
-    list_data = [[cell.value for cell in row] for row in sheet.rows()]
+    for sheet in workbook.sheets:
+        list_data = [[cell.value for cell in row] for row in sheet.rows()]
 
-    header = header_population(list_data[0])
-    data = data_population(list_data)
+        header = header_population(list_data[0])
+        data = data_population(list_data)
+        sheets[str(sheet)] = (header, data)
```
lyndsysimon: The keys for `sheets` get assigned here, but I don't see that they're ever used.

mfr/extensions/tabular/libs/ezodf_tools.py
```diff
@@ -12,11 +12,13 @@ def ods_ezodf(fp):
...
 
     return header, data
```
lyndsysimon: I understand that `header` and `data` refer to only the last sheet in the workbook - is that correct?

lyndsysimon:
First pass complete, pending response to the above questions.

mfr/extensions/tabular/libs/ezodf_tools.py
mfraezz: They're used to display the name of the sheet in the UI. See `viewer.mako`

mfr/extensions/tabular/libs/ezodf_tools.py
mfraezz: You're right, I should be returning `sheets`.

lyndsysimon:
Pass complete - it's all yours, Steve.

tests/extensions/tabular/test_panda_tools.py
```diff
@@ -9,24 +9,27 @@ class TestTabularPandaTools:
...
-            headers, data = panda_tools.csv_pandas(fp)
+            sheets = panda_tools.csv_pandas(fp)
 
-        assert headers[0] == {'field': 'one\ttwo\tthree', 'id': 'one\ttwo\tthree', 'name': 'one\ttwo\tthree', 'sortable': True}
-        assert data[0] == {'one\ttwo\tthree': 'a\tb\tc'}
+        sheet = sheets.popitem()[1]
+        assert sheet[0][0] == {'field': 'one\ttwo\tthree', 'id': 'one\ttwo\tthree', 'name': 'one\ttwo\tthree', 'sortable': True}
+        assert sheet[1][0] == {'one\ttwo\tthree': 'a\tb\tc'}
 
     # def test_dta_pandas():
```
HarryRybacki: @mfraezz do you know why this test is commented out / can we uncomment or remove it if it's just superfluous at this point?

mfr/extensions/tabular/static/js/bootstrap.min.js
```diff
@@ -0,0 +1,7 @@
+/*!
```
HarryRybacki: @mfraezz is bootstrap used anywhere else within the MFR and if is this needed?

HarryRybacki:
Pending minor clean up questions **CR2 is complete** :+1:

mfraezz:
Only the css. A slightly older version of bootstraps' css is being used for displaying certain error messages, but assuming PR #186 gets merged, otherwise no. 

I can refactor it to be globally usable, but my thought was that since this is the only extension using bootstrap.js, keeping it here until it's used by other extensions (if that happens) could slightly reduce bloat, as not every instance of the MFR needs to have every extension installed.

Also, seeing as how only two css classes are being used in the error messages, it might make sense to remove that dependency and use only those classes (similar to the rationale behind #186 ).

mfr/extensions/tabular/static/js/bootstrap.min.js
mfraezz: Only the css. A slightly older version of bootstraps' css is being used for displaying certain error messages, but assuming PR #186 gets merged, otherwise no. 

I can refactor it to be globally usable, but my thought was that since this is the only extension using bootstrap.js, keeping it here until it's used by other extensions (if that happens) could slightly reduce bloat, as not every instance of the MFR would need to have every extension installed.

Also, seeing as how only two css classes are being used in the error messages, it might make sense to remove that dependency and use only those classes (similar to the rationale behind #186 ).

tests/extensions/tabular/test_panda_tools.py
mfraezz: Uncertain. `.dta` isn't currently a supported filetype, judging from https://github.com/CenterForOpenScience/modular-file-renderer/blob/develop/setup.py#L647 being commented out. If support for it is re-enabled, this test should be used and updated. This wasn't directly associated with this PR, though, so I left it alone.

mfr/extensions/tabular/libs/xlrd_tools.py
```diff
@@ -17,29 +17,30 @@ def xlsx_xlrd(fp):
...
 
-    if sheet.ncols > max_size or sheet.nrows > max_size:
-        raise TableTooBigException("Table is too large to render.")
+    for sheet in wb.sheets():
+        if sheet.ncols > max_size or sheet.nrows > max_size:
+            raise TableTooBigException("Table is too large to render.")
 
-    if sheet.ncols < 1 or sheet.nrows < 1:
-        raise EmptyTableException("Table is empty or corrupt.")
+        if sheet.ncols < 1 or sheet.nrows < 1:
```
icereval: This exception needs to only be shown if no data is available for **ALL** sheets. In this current configuration having a single blank sheet which is **VERY** common in excel causes this to be raised thus the user never gets to see any workbook data.

mfraezz: Patched in #190

PR #188, opened 2015-08-31
Additions to .gitignore

jonathon-love:
following the build instructions leads to stuff being placed in the source tree which should not be committed. these are:
 - stuff from virtualenv
 - stuff from build dependencies

in this commit i add these to the .gitignore file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,4 +69,7 @@ player/files/*
 player/mfr_config_local.py
 player/static/mfr/*
 
+# virtualenv stuff
+include
+
 !.gitkeep
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ dist
 /build
 eggs
 parts
+src
 bin
 var
 sdist
@@ -37,6 +38,7 @@ develop-eggs
 .installed.cfg
 lib
 lib64
+share
 __pycache__
 .idea
 .hg


mfraezz:
It looks like your `virtualenvwrapper` might be incorrectly configured, those directories are normally found in `/.virtualenvs/[env_name]/`, which should be located in your home directory. Try:
```bash
> echo $VIRTUALENVWRAPPER_HOOK_DIR
```
to see where it is, and make sure you ran 
```bash
> mkvirtualenv --python `which python3` mfr
```
to make the environment. 

You might have done 
```bash
> virtualenv --python `which python3` mfr
```
which would tell `virtualenv` that you want the `mfr` directory to be the location for a virtualenv.

jonathon-love:
hi,

there is nothing assigned to my `$VIRTUALENVWRAPPER_HOOK_DIR` and my installation is a [very close to] vanilla ubuntu install.

> which would tell virtualenv that you want the mfr directory to be the location for a virtualenv.

perhaps you could update the documentation with this (or, even easier for your users, merge my commits :)

mfraezz:
There is actually a typo there, but the readme already says to run `mkvirtualenv`.

Could you tell me what the output is for: 
```bash
export | grep -i virtualenv
```

jonathon-love:
the of `export | grep -i virtualenv` output is blank!

and none of my VMs contain `mkvirtualenv` (i'd never heard of it before now).

where as the command

    virtualenv .

is well known to me... and works... except for there not being those entries in .gitignore :)

mfraezz:
Hmm.

`mkvirtualenv` is a function created by `virtualenvwrapper.sh`, which should be located in `/usr/local/bin/`. Do you have `virtualenvwrapper` installed?

jonathon-love:
ah, yes, that would be it. when i was first directed to mfr, i was sent to this page:

https://github.com/CenterForOpenScience/modular-file-renderer/blob/develop/CONTRIBUTING.rst

so i never saw the first page with the specific virtualenv instructions, and just did what i normally do.

with thanks

mfraezz:
I'll update that page too

PR #189, opened 2015-08-31
Added output at startup about address and port

jonathon-love:
I had to read the source code to figure out where the server was listening, and I thought this was a nice addition (and I've seen it done this way with quite a few server frameworks in the past).

Just a suggestion.


mfraezz:
Looks good :star: 

Thanks for submitting this

PR #190, opened 2015-09-04
[feature][fix][ready for review] Multisheet Ordering, Exception Handling, and UI

mfraezz:
Purpose
======
Patch bugs found in #187 after merge into develop

Changes
=======
* add an empty sheet to the set of sheets returned
* display empty sheet tab without data
* better logic for R/L arrow display
* use OrderedDict to maintain sheet ordering

Side Effects
=========
None

[#OSF-4302]

### DIFF
--- a/mfr/extensions/tabular/libs/xlrd_tools.py
+++ b/mfr/extensions/tabular/libs/xlrd_tools.py
@@ -1,6 +1,6 @@
 import xlrd
 from collections import OrderedDict
-from ..exceptions import TableTooBigException, EmptyTableException
+from ..exceptions import TableTooBigException
 
 from ..utilities import header_population
 from mfr.extensions.tabular.compat import range, basestring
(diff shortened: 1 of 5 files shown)


mfr/extensions/tabular/libs/xlrd_tools.py
```diff
@@ -43,4 +43,7 @@ def xlsx_xlrd(fp):
         header = header_population(fields)
         sheets[sheet.name] = (header, data)
 
+    if not sheets:
```
icereval: Sheets should be kept as an ordered dictionary as tab order is important in spreadsheets.

Also you might want to check with Erin if we should remove blank sheets or display them w/o any data (my preference would be the latter).

icereval: e.g. (for download) https://staging.osf.io/jw7gu/

PR #197, opened 2015-09-16
[Feature][Fix] Use case-insensitive search for tabular row filtering

mfraezz:
Purpose
=======
Search was case-sensitive, this can be annoying to users

Changes
=======
Use case insensitive RE matching. This was chosen over `.toUpperCase()` for special unicode characters.

Side Effects
=========
None

### DIFF
--- a/mfr/extensions/tabular/templates/viewer.mako
+++ b/mfr/extensions/tabular/templates/viewer.mako
@@ -78,7 +78,8 @@
                 var filtered = false;
                 var item = data[i];
                 for (var title in item) {
-                    if (search !== "" && item[title].toString().indexOf(search) !== -1) {
+                    var re = new RegExp(search, 'i');
+                    if (search !== "" && item[title].toString().match(re)) {
                         filtered = filtered || true;
                         continue;
                     }
--- a/mfr/extensions/tabular/templates/viewer.mako
+++ b/mfr/extensions/tabular/templates/viewer.mako
@@ -74,11 +74,11 @@
 
         function filterData(data, search) {
             var filteredData = [];
+            var re = new RegExp(search, 'i');
             for (var i = 0; i<data.length; i++){
                 var filtered = false;
                 var item = data[i];
                 for (var title in item) {
-                    var re = new RegExp(search, 'i');
                     if (search !== '' && item[title].toString().match(re)) {
                         filtered = filtered || true;
                         continue;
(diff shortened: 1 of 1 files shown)


abought:
The following information has been provided by @mfraezz to facilitate testing:

-  Set MFR to this branch, restart, and load a document in the OSF
- use a tabular file (e.g. csv, xlsx)
- you would probably need to restart the MFR server, yes
- in the search bar, type something that’s in one of the cells in any row, but use different capitalization
- if that row shows up, it’s working properly

mfr/extensions/tabular/templates/viewer.mako
```diff
@@ -78,7 +78,8 @@
                 var filtered = false;
                 var item = data[i];
                 for (var title in item) {
-                    if (search !== "" && item[title].toString().indexOf(search) !== -1) {
+                    var re = new RegExp(search, 'i');
+                    if (search !== "" && item[title].toString().match(re)) {
```
abought: Use single quotes for string constants (including empty string).
https://github.com/felixge/node-style-guide#use-single-quotes

(citation by way of [cosdev](http://cosdev.readthedocs.org/en/latest/style_guides/javascript.html))

abought:
CR1 complete. Confirmed it works for me- one small change requested to appease the style guide gods (in code that you inherited), but I'll move this ahead to CR2 meanwhile.

As a more general comment, you might also consider splitting `item[title].toString().match(re)` into a separate variable declaration for readability. Though comfort with multiple chained statements per line is a matter of personal preference.

samanehsan:
Done with CR2 and confirmed that this works for me as well :cat2: 

P.S. Thanks for the testing notes @abought and @mfraezz!

mfr/extensions/tabular/templates/viewer.mako
```diff
@@ -78,7 +78,8 @@
                 var filtered = false;
                 var item = data[i];
                 for (var title in item) {
-                    if (search !== "" && item[title].toString().indexOf(search) !== -1) {
+                    var re = new RegExp(search, 'i');
+                    if (search !== '' && item[title].toString().match(re)) {
```
sloria: Couldn't you just use `.toLower()` before comparing the strings?

mfraezz: Not if we want support for languages that use special unicode characters

mfr/extensions/tabular/templates/viewer.mako
```diff
@@ -78,7 +78,8 @@
                 var filtered = false;
                 var item = data[i];
                 for (var title in item) {
-                    if (search !== "" && item[title].toString().indexOf(search) !== -1) {
+                    var re = new RegExp(search, 'i');
```
sloria: This `RegExp` should only need to be created once, outside of the loop.

sloria:
Pass finished.

PR #199, opened 2015-10-06
[Feature Fix] Add wrap in pre on Firefox [OSF-3532]

haoyuchen1992:
## Purpose
In Firefox, `word-wrap: break-word` will not break the line in pre-tag.  To test it, you can run the [link](http://jsfiddle.net/oc49Lv3y/1/) and compare the result of firefox with any other major browsers.

resolve [OSF-3532]
## Change
There are two possible solutions. One is to add new pre css and apply it to all browsers -- [Link](https://css-tricks.com/snippets/css/make-pre-text-wrap/). But it need to apply all the other browser. And also it requires to override a lot of css on bootstrap.

So I tried another method that is to add firefox specific css on mfr.css.

 *  The tricks to target css on firefox  -- [Link](https://css-tricks.com/snippets/css/css-hacks-targeting-firefox/). Add code specific for Firefox
~~~
@-moz-document url-prefix() { 
  .selector {
     color:lime;
  }
}
~~~

 *  Use `white-space` and `word-break` to mimic `word-wrap` in firefox  and make it as the same as Chrome and other browsers. [link] http://stackoverflow.com/questions/4282757/how-to-make-word-break-on-firefox-using-css
~~~
/* For Firefox */
white-space: pre-wrap;
word-break: break-all;

/* For Chrome and IE */
word-wrap: break-word;
~~~

## Side Effect 
None

### DIFF
--- a/mfr/extensions/codepygments/static/css/default.css
+++ b/mfr/extensions/codepygments/static/css/default.css
@@ -61,3 +61,12 @@
 .highlight .vg { color: #19177C } /* Name.Variable.Global */
 .highlight .vi { color: #19177C } /* Name.Variable.Instance */
 .highlight .il { color: #666666 } /* Literal.Number.Integer.Long */
+
+/* CSS only for Firefox */
+@-moz-document url-prefix() {
+  .highlight > pre {
+      white-space: pre-wrap;       /*  css-3 */
+      white-space: -moz-pre-wrap;  /* Mozilla, since 1999 */
+      word-break: break-all;
+  }
+}
--- a/mfr/server/static/css/mfr.css
+++ b/mfr/server/static/css/mfr.css
@@ -70,12 +70,3 @@
 .embed-responsive-pdf {
     padding-bottom: 95%;
 }
-
-/* CSS only for Firefox */
-@-moz-document url-prefix() {
-  .highlight > pre {
-      white-space: pre-wrap;        /* css-3 */
-      white-space: -moz-pre-wrap;  /* Mozilla, since 1999 */
-      word-break: break-all;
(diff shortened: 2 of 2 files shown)


mfr/extensions/codepygments/static/css/default.css
```diff
@@ -60,4 +60,13 @@
...
+.highlight .il { color: #666666 } /* Literal.Number.Integer.Long */
+
+/* CSS only for Firefox */
+@-moz-document url-prefix() {
+  .highlight > pre {
+      white-space: pre-wrap;       /*  css-3 */
+      white-space: -moz-pre-wrap;  /* Mozilla, since 1999 */
+      word-break: break-all;
+  }
+}
```
abought: Incredibly minor, but generally try to make sure that all files end with a newline. (some IDEs have a setting to check this automatically)

It won't make or break the OSF, but every now and then a tool will work oddly if there's no newline at the end of a file. (hence why github shows that warning icon) See: http://stackoverflow.com/questions/729692/why-should-files-end-with-a-newline

abought:
CR1 complete; no changes requested.

Confirmed that this change works as intended when running locally; FF and Chrome.

Note to reviewers: to view this locally, make sure that waterbutler is updated to work with the newest osf develop branch. (changes landed earlier today)

chennan47:
CR2 complete. Looks good.

mfr/extensions/codepygments/static/css/default.css
haoyuchen1992: @abought Thanks for the resource. I have never thought about it before!

PR #211, opened 2016-04-29
[CR] Convert MFR to use python-3.5

felliott:
Continuing @chrisseto's work, finish the 3.5 conversion of MFR.

Notes:
* 3.5-based MFR will only work with a 3.5-based WB, so the WB dep has been bumped to v0.19.0.
* The server `DEBUG` must be set to False because of [this bug in core python](https://bugs.python.org/issue25394).  The default has been updated to False, so you should only need to change anything if you have set it in `~/.cos/mfr-test.json`.
* `pandas` and `rpy2` were bumped to take advantage of upstream 3.5 fixes.
* a couple of hacks needed for 3.4 were removed.

Fixes: [#OSF-6143]

### DIFF
--- a/mfr/providers/osf/provider.py
+++ b/mfr/providers/osf/provider.py
@@ -111,4 +111,4 @@ class OsfProvider(provider.BaseProvider):
         if self.authorization:
             kwargs.setdefault('headers', {})['Authorization'] = 'Bearer ' + self.token
 
-        return (await aiohttp.request(method, url, *args, **kwargs))
+        return await aiohttp.request(method, url, *args, **kwargs)
--- a/mfr/providers/osf/provider.py
+++ b/mfr/providers/osf/provider.py
@@ -1,7 +1,7 @@
 import os
 import json
-import asyncio
 import hashlib
+import logging
 import mimetypes
 
 import furl
@@ -11,6 +11,8 @@ from waterbutler.core import streams
 
 from mfr.core import exceptions
 from mfr.core import provider
+
+logger = logging.getLogger(__name__)
 
 
 class OsfProvider(provider.BaseProvider):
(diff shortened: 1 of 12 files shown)


mfr/providers/osf/provider.py
```diff
@@ -99,4 +111,4 @@ def _make_request(self, method, url, *args, **kwargs):
         if self.authorization:
             kwargs.setdefault('headers', {})['Authorization'] = 'Bearer ' + self.token
 
-        return (yield from aiohttp.request(method, url, *args, **kwargs))
+        return (await aiohttp.request(method, url, *args, **kwargs))
```
chrisseto: Extra parens are no longer needed

mfr/server/handlers/core.py
```diff
@@ -113,16 +120,15 @@ def initialize(self):
...
             super().initialize(self.modules[module_name])
-            return (yield super().get(path))
+            return (await super().get(path))
         except Exception:
             self.set_status(404)
 
         try:
             super().initialize(settings.STATIC_PATH)
-            return (yield super().get(path))
+            return (await super().get(path))
```
chrisseto: More unnecessary parens

mfr/server/handlers/export.py
```diff
@@ -59,20 +55,27 @@ def get(self):
...
+            await self.write_stream(waterbutler.core.streams.FileStreamReader(fp))
 
-    @tornado.gen.coroutine
     def on_finish(self):
         if self.request.method not in self.ALLOWED_METHODS:
             return
 
+        # Spin off upload into non-blocking operation
+        loop = asyncio.get_event_loop()
+        loop.call_soon(
```
chrisseto: Wouldn't `asyncio.ensure_future(self._cache_and_clean())` do the same thing?

felliott: I did that first, but noticed that render.py [is doing it this way](https://github.com/CenterForOpenScience/modular-file-renderer/blob/feature/python-3.5/mfr/server/handlers/render.py#L61).  Is there a reason for that?

chrisseto: I'm not sure...

felliott:
Merged.

PR #212, opened 2016-05-11
Sort stringified numbers as numbers, not strings

mfraezz:
Purpose
=======
The compare function that sorts tabular data should but does not compare stringified numbers properly.

screen shot 2016-05-11 at 10 54 03

Changes
=======
* If values to be compared are numbers, parse them as such. Otherwise, use default string ordering.

screen shot 2016-05-11 at 10 55 24

Side Effects
==========
Columns populated with both strings and numbers still behave oddly, but it's no worse than before.

screen shot 2016-05-11 at 11 00 32
screen shot 2016-05-11 at 11 00 39
screen shot 2016-05-11 at 11 00 52
screen shot 2016-05-11 at 11 01 00

### DIFF
--- a/mfr/extensions/tabular/templates/viewer.mako
+++ b/mfr/extensions/tabular/templates/viewer.mako
@@ -91,6 +91,15 @@
             return filteredData;
         }
 
+        function prepareSortValue(s) {
+            // Everything retrieved from SlickGrid is a string.
+            // If those strings are numbers, parse them.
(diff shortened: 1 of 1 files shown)


felliott:
LGTM!  I've added a commit to factor out the numish testing and fix a minor style issue (not yours).  Is there a ticket for this I can point QA to?

mfraezz:
> Is there a ticket for this I can point QA to?

Nope. I noticed it and it bugged me, was a quick fix.

PR #213, opened 2016-05-18
Line wrap [OSF-5875]

zamattiac:
## Purpose

When a .txt file is uploaded in Safari, the line overflow will wrap around when rendered there, but the same file will have to scroll to the right in Chrome.

image

## Similar problems

Firefox also had this problem [OSF-3532] and it was resolved with a Firefox-specific line in the CSS.

## Changes

Added a line to the CSS in mfr/extensions/codepygements/static/css/default.css for the pre element:

`  white-space : pre-line;   ` 

so it wraps to the next line now.

image

Removed Firefox-specific code to support Chrome as well.

## Side Effects
- Safari: Words no longer break to wrap in the middle of the word
- Firefox: Words no longer break to wrap in the middle of the word
- CSS simplification

### DIFF
--- a/mfr/extensions/codepygments/static/css/default.css
+++ b/mfr/extensions/codepygments/static/css/default.css
@@ -82,11 +82,12 @@ code, kbd, pre, samp {
     box-sizing: border-box;
 }
 
-/* CSS only for Firefox */
+/* CSS only for Firefox
 @-moz-document url-prefix() {
   .highlight > pre {
-      white-space: pre-wrap;       /*  css-3 */
-      white-space: -moz-pre-wrap;  /* Mozilla, since 1999 */
+      white-space: pre-wrap;
+      white-space: -moz-pre-wrap;
       word-break: break-all;
   }
 }
+*/
--- a/mfr/extensions/codepygments/static/css/default.css
+++ b/mfr/extensions/codepygments/static/css/default.css
@@ -71,7 +71,7 @@ pre {
     line-height: 1.42857;
     background-color: #F5F5F5;
     border: 1px solid #CCC;
-    white-space:pre-line;
+    white-space: pre-line;
 }
 
 code, kbd, pre, samp {
@@ -81,13 +81,3 @@ code, kbd, pre, samp {
 * {
     box-sizing: border-box;
 }
-
-/* CSS only for Firefox
-@-moz-document url-prefix() {
-  .highlight > pre {
-      white-space: pre-wrap;
-      white-space: -moz-pre-wrap;
-      word-break: break-all;
-  }
-}
-*/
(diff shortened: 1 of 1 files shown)


mfr/extensions/codepygments/static/css/default.css
```diff
@@ -71,6 +71,7 @@ pre {
     line-height: 1.42857;
     background-color: #F5F5F5;
     border: 1px solid #CCC;
+    white-space:pre-line;
```
felliott: Why use `pre-line` instead of `pre-wrap`?  (Genuine question, not sure myself.)

mfr/extensions/codepygments/static/css/default.css
```diff
@@ -81,11 +82,12 @@ code, kbd, pre, samp {
...
 @-moz-document url-prefix() {
   .highlight > pre {
-      white-space: pre-wrap;       /*  css-3 */
-      white-space: -moz-pre-wrap;  /* Mozilla, since 1999 */
+      white-space: pre-wrap;
+      white-space: -moz-pre-wrap;
       word-break: break-all;
   }
 }
+*/
```
felliott: Instead of commenting this out, please delete it.  We'll still have it in the git history.

mfr/extensions/codepygments/static/css/default.css
felliott: (minor) please put a space after the `:`.

felliott:
Sending back to 'Additional Development', but only minor changes needed.

zamattiac:
@felliott Thanks, and [this website](http://www.impressivewebs.com/css-white-space/) explains it pretty well

felliott:
> @felliott <https://github.com/felliott> Thanks, and this website <http://www.impressivewebs.com/css-white-space/> explains it pretty well

Thanks for the link!  After reading that, I think we do want `pre-wrap` instead.  A lot of folks use whitespace for formatting in their .txt files, and we wouldn’t want to throw those off.  An ascii table would look pretty weird if you compressed down the white space.

zamattiac:
Okay I will fix this in the next five minutes

On Thu, May 19, 2016 at 9:36 AM, Fitz Elliott <notifications@github.com>
wrote:

> > @felliott <https://github.com/felliott> Thanks, and this website <
> http://www.impressivewebs.com/css-white-space/> explains it pretty well
>
> Thanks for the link! After reading that, I think we do want `pre-wrap`
> instead. A lot of folks use whitespace for formatting in their .txt files,
> and we wouldn’t want to throw those off. An ascii table would look pretty
> weird if you compressed down the white space.
>
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/CenterForOpenScience/modular-file-renderer/pull/213#issuecomment-220326140>
>



-- 
Matthew Keitelman
571-269-4724
mak2vr@virginia.edu

felliott:
LGTM, thanks, @zamattiac!  This will be merged after the next MFR release.

felliott:
QA had some spare bandwidth, so this has been merged, tested, and will be part of the next MFR release!  Closing.

PR #214, opened 2016-05-19
[feature] Should render LaTeX files in MFR [OSF-6331]

acshi:
## Purpose

Render LaTeX files in MFR instead of just showing them with syntax highlighting.

Before:
before

After:
after
## Changes

The exporter and renderer for unoconv is largely copied and then modified to use pdflatex instead of unoconv. The result is that LaTeX files can be rendered and shown as pdfs. However, images or other embedded/included content cannot be shown. A more complicated process would be required to have these assets included that would involve coupling MFR with waterbutler/osf.

In order for this to work, pdflatex needs to be installed. Unfortunately, the default installation of MacTex is about 2.5GB large. On the other hand, the stripped down 90MB BasicTex is severely lacking in functionality. There doesn't seem to be an in-between version that just lacks the unnecessary authoring tools that come in the full installation. 

MacTec can be installed with "brew cask install mactex" after brew-cask is installed with "brew install brew-cask".

If pdflatex is not found on the path, its location should be added to the MFR settings under:
LATEX_EXTENSION_CONFIG.PDFLATEX_BIN
## Side Effects

The only possible side effects might be a result of installing MacTex/pdflatex.
## Ticket

https://openscience.atlassian.net/browse/OSF-6331

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,10 @@
 ChangeLog
 *********
 
+0.18.3 (2017-01-11)
+===================
+- Fix: Increase max codepygments render size to 200kb from 64kb.
+
 0.18.2 (2017-01-04)
 ===================
 - Happy New Year!
--- a/mfr/__init__.py
+++ b/mfr/__init__.py
@@ -1,2 +1,2 @@
-__version__ = '0.18.2'
+__version__ = '0.18.3'
 __import__('pkg_resources').declare_namespace(__name__)
--- a/mfr/extensions/codepygments/settings.py
+++ b/mfr/extensions/codepygments/settings.py
@@ -2,4 +2,4 @@ from mfr import settings
 
 config = settings.child('CODEPYGMENTS_EXTENSION_CONFIG')
 
-MAX_SIZE = int(config.get('MAX_SIZE', 65536))
+MAX_SIZE = int(config.get('MAX_SIZE', 204800))  # 200kb


felliott:
Thanks, @acshi!  Lemme summon our DevOps team to weigh in on adding a 2.5GB dependency.

@mattclark, @icereval: I beckon thee! Is there any reason to avoid installing a latex package in the Dockerfile?  If not, I'm assuming you'll need a list of packages to add to the `apt-get` invocation, yes?

acshi:
If we are using apt-get, the full installation can be done as apt-get install texlive-full which I hear is about 1.8GB because it doesn't include some of the things included in MacTex. Alternatively, a more basic installation could be done with just the packages latex-recommended, latex-extra, fonts-recommended, fonts-extra and their dependencies, and should take less than 500MB. However, it is hard to tell how many of the more obscure packages would occur in user latex files.

mattclark:
This is possible, however there are security concerns associated with LaTex that we would need to  investigate and test. It would be a significant undertaking.

whit537:
Doing some digging ...

> By default latex can not execute system code or write files beginning `.` or write out of the current directory, so the security risks are minimal.

http://tex.stackexchange.com/questions/262625/security-latex-injection-hack

Hmm ...

whit537:
> the articles tagged with security

http://tex.stackexchange.com/questions/tagged/security

whit537:
"[Don’t take LaTeX files from strangers](http://c59951.r51.cf2.rackcdn.com/5670-73506-checkoway.pdf)"

acshi:
"A web-based TEX previewer: the ecstasy and the agony"

https://tug.org/tug2010/booklet/doob/preprint.pdf

whit537:
Poring over the TeX StackExchange additionally turns up:

"[Don't take LaTeX files from strangers](http://cseweb.ucsd.edu/~hovav/dist/tex-login.pdf)" (original; [h/t](http://tex.stackexchange.com/a/87039))

"[A web-based TeX previewer: The ecstasy and the agony](http://www.tug.org/TUGboat/tb31-2/tb98doob.pdf)" ([h/t](http://tex.stackexchange.com/questions/10418/how-can-i-safely-compile-other-peoples-latex-documents#comment18053_10418))

whit537:
> "A web-based TEX previewer: the ecstasy and the agony"

👍

whit537:
This may also be included in "Don't take LaTeX files from strangers":

"[Are Text-Only Data Formats Safe? Or, Use This LaTeX Class File to Pwn Your Computer](http://cseweb.ucsd.edu/~hovav/dist/texhack.pdf)" ([h/t](http://tex.stackexchange.com/questions/4589/what-analysis-of-texlives-restricted-permissions-model-exists#comment6807_4589))

whit537:
I found two classes of SO threads that were interesting (filtering for TeX Live since that I [gather](https://github.com/CenterForOpenScience/modular-file-renderer/pull/214#issuecomment-220611902) that's what we'd use in production):

 1. Questions about the security of TeX packaging systems:
  - "[How do CTAN maintainers make sure all the uploaded packages and classess in CTAN are free of malicious code?](http://tex.stackexchange.com/questions/103056/how-do-ctan-maintainers-make-sure-all-the-uploaded-packages-and-classess-in-ctan)"
  - "[Full version of TeX Live: malware checked?](http://tex.stackexchange.com/questions/134656/full-version-of-tex-live-malware-checked)"
  - "[Tex Live Package Security](http://tex.stackexchange.com/questions/279044/tex-live-package-security)"
  - "[Is CTAN secure against attackers uploading bugged versions of well-known packages?](http://tex.stackexchange.com/questions/305556/is-ctan-secure-against-attackers-uploading-bugged-versions-of-well-known-package)"

 2. Questions about vulnerabilities in TeX itself:
  - "[What analysis of Texlive's restricted permissions model exists?](http://tex.stackexchange.com/questions/4589/what-analysis-of-texlives-restricted-permissions-model-exists)"
  - "[How can I safely compile other people's LaTeX documents?](http://tex.stackexchange.com/questions/10418/how-can-i-safely-compile-other-peoples-latex-documents)"
  - "[Disable including of system files [duplicate]](http://tex.stackexchange.com/questions/58311/disable-including-of-system-files)"
  - "[calling latex from php: security issues](http://tex.stackexchange.com/questions/87038/calling-latex-from-php-security-issues)"
  - "[Is Arara safe/secure](http://tex.stackexchange.com/questions/103466/is-arara-safe-secure)"
  - "[Security: Latex Injection Hack](http://tex.stackexchange.com/questions/262625/security-latex-injection-hack)"

whit537:
Suggestions from "A web-based TeX previewer":

 - sanitize any log output that is displayed to the user (2.1)
 - constrain resources to avoid denial of service (2.2)
 - require authentication (2.2.3)

whit537:
From djb's "[Some thoughts on security after ten years of qmail 1.0](http://cr.yp.to/qmail/qmailsec-20071101.pdf)" (via "Are Text-Only Data Formats Safe?"):

- Prohibit new files, new sockets, etc., by setting the current and maximum `RLIMIT_NOFILE` limits to 0.
- Prohibit filesystem access: `chdir` and `chroot` to an empty directory.
- Choose a uid dedicated to this process ID. This can be as simple as adding the process ID to a base uid, as long as other system-administration tools stay away
from the same uid range.
- Ensure that nothing is running under the uid: `fork` a child to run `setuid(targetuid)`, `kill(-1,SIGKILL)`, and `_exit(0)`, and then check that the child exited normally.
- Prohibit `kill()`, `ptrace()`, etc., by setting gid and uid
to the target uid.
- Prohibit `fork()`, by setting the current and maximum `RLIMIT_NPROC` limits to 0.
- Set the desired limits on memory allocation and other resource allocation.

whit537:
Maybe https://github.com/openjudge/sandbox is helpful?

whit537:
From "Are Text-Only Data Formats Safe?":

- audit the interpreter
- establish a safe subset of commands
- sandbox it (**recommended**)

whit537:
Skimming the two versions of "Don’t take LaTeX files from strangers" suggest that they are functionally equivalent to each other (and largely derived from "Are Text-Only Data Formats Safe?"). I'm going to read the [nicer-looking one](http://c59951.r51.cf2.rackcdn.com/5670-73506-checkoway.pdf). :-)

whit537:
Yeah, nothing new there, really.

whit537:
Okay! So [MathTran](http://mathtran.open.ac.uk/) is the most secure option reviewed ... are they open-source? We should also look into https://github.com/sharelatex/sharelatex.

whit537:
To be continued ...

whit537:
Alright, blocking this on adding a sandboxer to MFR ([OSF-6547](https://openscience.atlassian.net/browse/OSF-6547)).

PR #215, opened 2016-05-24
Feature/Handle case when waterbutler v1 url provided.

pattisdr:
Needed for https://openscience.atlassian.net/browse/EOSF-19

# Purpose

MFR was written before APIv2.   If you give it a url with a file GUID in it, it then attempts to fetch the download url.  However, the APIv2 already returns a download url for a file. 

# Changes
Handle the case when the download url is passed directly to MFR.

### DIFF
--- a/mfr/providers/osf/provider.py
+++ b/mfr/providers/osf/provider.py
@@ -84,7 +84,7 @@ class OsfProvider(provider.BaseProvider):
         return streams.ResponseStreamReader(response, unsizable=True)
 
     async def _fetch_download_url(self):
-        # Waterbutler url provided
+        # v1 Waterbutler url provided
         if 'v1/resources' in self.url:
             return self.url
         if not self.download_url:
--- a/mfr/providers/osf/provider.py
+++ b/mfr/providers/osf/provider.py
@@ -2,6 +2,7 @@ import os
 import json
 import hashlib
 import logging
+from urllib.parse import urlparse
 import mimetypes
 
 import furl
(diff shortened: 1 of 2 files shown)


mfr/providers/osf/provider.py
```diff
@@ -84,6 +84,9 @@ def __init__(self, request, url):
         return streams.ResponseStreamReader(response, unsizable=True)
 
     async def _fetch_download_url(self):
+        # Waterbutler url provided
```
felliott: Minor: can you update the comment to specify that we're looking for a v1 WaterButler url?

mfr/providers/osf/provider.py
```diff
@@ -84,6 +84,9 @@ def __init__(self, request, url):
...
+        if 'v1/resources' in self.url:
```
felliott: I think this will pass if we're trying to download a file with `v1/resources` in the filename by the v0 api (possible with non-osfstorage).  e.g. `http://localhost:7777/file?path=/foo/v1/resource-center.txt&provider=googledrive&nid=vjsdy&accept_url=false&mode=render`.  Can you anchor the match to start at the beginning of the url?

pattisdr: I changed it to check if /v1/resources/ appeared in the beginning of the path.

felliott:
This looks good to me.  I'll merge it into develop after the next MFR release.

samchrisinger:
@felliott any ETA on getting this on staging?

felliott:
It'll be up after the next release of MFR.  I'm just waiting for QA to finish verifying the 3.5 conversion.

felliott:
Merged into develop.

PR #217, opened 2016-07-13
Limit codePygmentRenderer to 64k files

TomBaxter:
## Purpose:
[OSF-6670](https://openscience.atlassian.net/browse/OSF-6670)
Large text file are failing to renderer in a disorderly manner.
Limit them.

## Changes:
Add mfr/extensions/codepygments/exceptions.py add FileToLargeException
Update mfr/extensions/codepygments/render.py check size and raise FileToLargeException
Update mfr/extensions/codepygments/settings.py add MAX_SIZE configuration
Update  requirements.txt add humanfriendly requirement
Update tests/extensions/codepygments/test_renderer.py add tests
1. test_render_codepygments_max_size
2. test_render_codepygments_over_size

## Side effects
This PR is lock step with OSF PR#5969 limiting the size of text file that we'll allow to use the share editor. If these limits get out of sync, users may be confused.

[#OSF-6670]


mfr/extensions/codepygments/render.py
```diff
@@ -20,6 +21,9 @@ class CodePygmentsRenderer(extension.BaseRenderer):
         ]).get_template('viewer.mako')
 
     def render(self):
+        file_size = os.path.getsize(self.file_path)
+        if file_size > 65536:
```
mfraezz: Could this be made configurable, similarly to how `tabular/settings.py` are set?

TomBaxter: Complete. Also merge conflicts fixed. Travis build of the beast passed.

mfr/extensions/codepygments/render.py
```diff
@@ -25,6 +27,9 @@ def __init__(self, *args, **kwargs):
         self.metrics.add('pygments_version', pygments.__version__)
 
     def render(self):
+        file_size = os.path.getsize(self.file_path)
+        if file_size > settings.MAX_SIZE:
+            raise(cp_exceptions.FileToLargeException('Text files larger than 64kb are not rendered; please download the file to view.'))
```
mfraezz: My only concern with this is that if we decide to change `MAX_SIZE`, this figure won't be accurate. 

Not blocking, but consider using some util to format `MAX_SIZE`. I don't know of any built-ins that do this, but something like this would work:

```
In [1]: %paste
def stringify_bytecount(b):
    i = 0
    sz = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
    while b >= 1024 and i < len(sz)-1:
        i += 1
        b /= 1024.0
    return '{} {}'.format(round(b, 1), sz[i]).replace('.0 ', ' ')

## -- End pasted text --

In [2]: stringify_bytecount(6553300000000000000000000000)
Out[2]: '5420.8 YB'

In [3]: stringify_bytecount(65536)
Out[3]: '64 KB'
```

mfraezz: Update: There's a library called `humanfriendly` that does this, and WaterButler's already using it:
```
In [1]: import humanfriendly

In [2]: humanfriendly.format_size??
Signature: humanfriendly.format_size(num_bytes, keep_width=False)
Source:
def format_size(num_bytes, keep_width=False):
    """
    Format a byte count as a human readable file size (supports ranges from
    kilobytes to terabytes).

    :param num_bytes: The size to format in bytes (an integer).
    :param keep_width: ``True`` if trailing zeros should not be stripped,
                       ``False`` if they can be stripped.
    :returns: The corresponding human readable file size (a string).

    Some examples:

    >>> from humanfriendly import format_size
    >>> format_size(0)
    '0 bytes'
    >>> format_size(1)
    '1 byte'
    >>> format_size(5)
    '5 bytes'
    >>> format_size(1024 ** 2)
    '1 MB'
    >>> format_size(1024 ** 3 * 4)
    '4 GB'
    """
    for unit in reversed(disk_size_units):
        if num_bytes >= unit['divider']:
            number = round_number(float(num_bytes) / unit['divider'], keep_width=keep_width)
            return pluralize(number, unit['singular'], unit['plural'])
    return pluralize(num_bytes, 'byte')
File:      ~/.virtualenvs/mss35/lib/python3.5/site-packages/humanfriendly/__init__.py
Type:      function

In [3]: humanfriendly.format_size(6553300000000000000000000000)
Out[3]: '5820499637820.83 PB'

In [4]: humanfriendly.format_size(65536)
Out[4]: '64 KB'
```

TomBaxter: Updated to use humanfriendly and added two tests.

felliott:
Merged!  Thank you!

PR #218, opened 2016-07-28
MFR - validate input and throw 400

TomBaxter:
## Purpose:
[OSF-6789](https://openscience.atlassian.net/browse/OSF-6789)
Improve error handling to improve errors that get to Sentry

## Changes:
Updated mfr/extensions/codepygments/render.py Update def _render_html to catch encoding None
Updated mfr/providers/osf/provider.py def __init__ to catch absent x-waterbutler-metadata header
Updated mfr/server/handlers/core.py class BaseHandler to catch absent url argument

## Caveat:
I was unable to find a way to exercise this code.

[#OSF-6789]

### DIFF
--- a/mfr/extensions/codepygments/render.py
+++ b/mfr/extensions/codepygments/render.py
@@ -49,10 +49,11 @@ class CodePygmentsRenderer(extension.BaseRenderer):
 
         if exception is not None:
             encoding = chardet.detect(data)
-            try:
-                content = data.decode(encoding['encoding'])
-            except UnicodeDecodeError as e:
-                exception = e
+            if encoding['encoding']:
+                try:
+                    content = data.decode(encoding['encoding'])
+                except UnicodeDecodeError as e:
+                    exception = e
 
         if content is None:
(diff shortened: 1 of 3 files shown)


mfr/extensions/codepygments/render.py
```diff
@@ -49,10 +49,11 @@ def _render_html(self, fp, ext, *args, **kwargs):
...
             encoding = chardet.detect(data)
-            try:
-                content = data.decode(encoding['encoding'])
-            except UnicodeDecodeError as e:
-                exception = e
+            if encoding['encoding']:
+                try:
+                    content = data.decode(encoding['encoding'])
+                except UnicodeDecodeError as e:
+                    exception = e
```
mfraezz: Should an error be raised here `if not encoding['encoding']`? 

I know `AssertionError('Got no content or exception')` will probably be raised, but that's not exactly indicative of what the actual error was.

felliott:
Thanks, @TomBaxter!  Applied with minor changes.

PR #223, opened 2016-10-02
[Feature] Environment Variable Settings Support

icereval:
Support for settings overrides w/ Environment Variables for simple values.

[Twelve-Factor App Config](https://12factor.net/config)

Discovered need when simplifying docker-compose integration w/ OSF.

#### Deployment Considerations
- [ ] SERVER_CONFIG  -> ALLOWED_PROVIDER_DOMAINS changed from a List to Literal String of space separated domains.

  e.g. `'http://localhost:5000/ http://localhost:7777/'`

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -2,13 +2,10 @@
 
 -e git+https://github.com/centerforopenscience/aiohttpretty.git@0.0.2#egg=aiohttpretty
 colorlog==2.5.0
-flake8==2.3.0
+flake8==3.0.4
 ipdb
 mccabe
-pep8
 pyflakes
 pytest==2.8.2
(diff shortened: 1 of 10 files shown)


felliott:
How do you want to handle booleans and legitimate `None`s?  I've got some local code to infer booleans from a limited list of strings, and die screaming otherwise.  I'm less confident about stringy `None`s.  We could make `'None'` a magic value or update the code to not use default 'None' config vals.  Do you have a preference?  `int`s I'm just straight casting in the config files.

icereval:
Booleans are typically handled by setting the environment literal value to a blank value, e.g.

`DEBUG=`

(nothing after the `=`)

felliott:
Merged into develop.

PR #224, opened 2016-10-24
Stuff error message to prevent IE from consuming it

alexschiller:
## Purpose
Internet explorer consumes error messages that are smaller than 512 bytes

## Changes

This PR stuffs the error message to be slightly larger, preventing IE from throwing its custom 404 page error.

## Side Effects
None.

## Jira Ticket
https://openscience.atlassian.net/browse/OSF-7040


felliott:
Merged into develop as 334a3da0f784c443adaae1579fb7c605c9df6a8d.  Thanks, @alexschiller!

PR #230, opened 2017-01-24
[SVCS-123] Allows read-the-docs to read Python 3

Johnetordoff:
# Purpose

Upgrade invoke to 0.13.0 and write a work around so mfr can use python 3 with read-the-docs.

# Changes

Additional config files for read the docs, removes references to non-existent LaTex files.

# Side Effects

None that I know of.

# Ticket

https://openscience.atlassian.net/browse/SVCS-123


felliott:
Merged as part of b7e9f1dfb370e5e044e16bbfb467cae6cf2f4f8a.  Thanks!

PR #235, opened 2017-01-27
[SVCS-154] Limit jsc3d zoom factor

Johnetordoff:
# Purpose

User can zooming out indefinitely and lose track of rendered object, this prevents that.

# Changes 

Small js fix prevents zoom out to a ludicrous distance.

# Side Effects

None that I know of.

# Ticket

https://openscience.atlassian.net/browse/SVCS-154


Johnetordoff:
@felliott It's kind of hard to explain without having a visual aid but you shouldn't limit zooming because the system also supports panning so being too close to an object isn't preventable by limiting the "zoomFactor" in the same way as zooming out. Also it would be very problematic to the view the insides of hollow objects with a zoom limit.

felliott:
Merged as e8f890f9f8da701c564528894aa1074be33ef175.

PR #238, opened 2017-02-02
[SVCS-241][SVCS-53] Add maximum file size limit for rendering files with certain extensions.

Johnetordoff:
# Purpose

Add maximum file size limit for rendering files with certain extensions.

# Changes

Adds new type of exception and a throws it when metadata indicates a file has exceeded the maximum size for it's extension.

# Side Effects

None that I know of

# Ticket

https://openscience.atlassian.net/browse/SVCS-241

Related to:

https://openscience.atlassian.net/browse/SVCS-53
https://openscience.atlassian.net/browse/SVCS-235

Figure (New Behavior):
screen shot 2017-02-02 at 10 10 19 am

### DIFF
--- a/mfr/extensions/tabular/exceptions.py
+++ b/mfr/extensions/tabular/exceptions.py
@@ -33,14 +33,6 @@ class TableTooBigError(TabularRendererError):
         super().__init__(message, *args, code=code, **kwargs)
         self.attr_stack.append([self.__TYPE, {}])
 
-class TooBigToRenderError(TabularRendererError):
-
-    __TYPE = 'tabular_too_big_to_render'
-
-    def __init__(self, message, *args, code: int=400, **kwargs):
-        super().__init__(message, *args, code=code, **kwargs)
-        self.attr_stack.append([self.__TYPE, {}])
-
 
 class UnexpectedFormattingError(TabularRendererError):
 
--- a/mfr/extensions/tabular/exceptions.py
+++ b/mfr/extensions/tabular/exceptions.py
@@ -33,6 +33,14 @@ class TableTooBigError(TabularRendererError):
         super().__init__(message, *args, code=code, **kwargs)
         self.attr_stack.append([self.__TYPE, {}])
 
+class TooBigToRenderError(TabularRendererError):
+
+    __TYPE = 'tabular_too_big_to_render'
+
+    def __init__(self, message, *args, code: int=400, **kwargs):
+        super().__init__(message, *args, code=code, **kwargs)
+        self.attr_stack.append([self.__TYPE, {}])
+
 
 class UnexpectedFormattingError(TabularRendererError):
 
--- a/mfr/providers/osf/provider.py
+++ b/mfr/providers/osf/provider.py
@@ -100,7 +100,7 @@ class OsfProvider(provider.BaseProvider):
 
         name, ext = os.path.splitext(metadata['data']['name'])
 
-        if  MAX_FILE_SIZE_TO_RENDER.get(ext) and metadata['data']['size'] > MAX_FILE_SIZE_TO_RENDER.get(ext):
+        if MAX_FILE_SIZE_TO_RENDER.get(ext) and metadata['data']['size'] > MAX_FILE_SIZE_TO_RENDER.get(ext):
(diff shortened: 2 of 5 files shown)


mfr/settings.py
```diff
@@ -84,6 +84,9 @@ def child(self, key):
 UNSUPPORTED_EXPORTER_MSG = 'Exporting of this file type is not currently supported.'
 UNSUPPORTED_RENDER_MSG = 'Viewing of this file type is not currently supported. Please download the file to view.'
 
+
+MAX_FILE_SIZE_TO_RENDER = {'csv': 100000, '.xlsx': 100000, '.tsv': 100000, '.sav': 100000, '.xls': 100000}
```
felliott: I think `csv` should be `.csv`.  Also, I think the ticket said to start with 100Mb, but this is 100Kb, yes?

mfr/providers/osf/provider.py
```diff
@@ -96,6 +99,12 @@ def __init__(self, request, url):
         # }}
 
         name, ext = os.path.splitext(metadata['data']['name'])
+
+        if MAX_FILE_SIZE_TO_RENDER.get(ext) and metadata['data']['size'] > MAX_FILE_SIZE_TO_RENDER.get(ext):
```
felliott: Instead of calling `MAX_FILE_SIZE_TO_RENDER.get(ext)` 2 or 3 times, go ahead and save it to an intermediate variable.

felliott: I believe some providers may have  `metadata['data']['size'] == None`, which would probably throw a type error here.  If size is unknown, let's skip this check.

mfr/providers/osf/provider.py
```diff
@@ -96,6 +99,12 @@ def __init__(self, request, url):
...
+            size = sizeof_fmt(MAX_FILE_SIZE_TO_RENDER.get(ext))
+            raise TooBigToRenderError("This file with extension '{ext}' exceeds the size limit of {size} and will not "
```
felliott: Max line length for MFR should be 100 chars, same as WaterButler.  I thought this was in the CONTRIBUTING, but I'm not seeing it.  Boo on me.

mfr/providers/osf/provider.py
```diff
@@ -96,6 +99,12 @@ def __init__(self, request, url):
...
+                                      "be rendered. To view this file download it and view it offline.".format(**locals()))
```
felliott: I'd prefer this to be explicit, either with `.format(ext=ext, size=size)` or positionals.  I'm squeamish about possibly exposing local private variables in an error message.

mfr/providers/osf/provider.py
```diff
@@ -13,7 +13,10 @@
 
 from mfr.core import exceptions
 from mfr.core import provider
+from mfr.core.utils import sizeof_fmt
 from mfr.providers.osf import settings
+from mfr.settings import MAX_FILE_SIZE_TO_RENDER
+from mfr.extensions.tabular.exceptions import TooBigToRenderError
```
felliott: Validating in the provider is a good idea, but in that case this error should be a TabualrError.  Make it a subclass of ProviderError instead.

mfr/extensions/tabular/exceptions.py
```diff
@@ -33,6 +33,14 @@ def __init__(self, message, *args, code: int=400, **kwargs):
         super().__init__(message, *args, code=code, **kwargs)
         self.attr_stack.append([self.__TYPE, {}])
 
+class TooBigToRenderError(TabularRendererError):
+
+    __TYPE = 'tabular_too_big_to_render'
+
+    def __init__(self, message, *args, code: int=400, **kwargs):
+        super().__init__(message, *args, code=code, **kwargs)
+        self.attr_stack.append([self.__TYPE, {}])
```
felliott: One bit of data that would be useful to track for the metrics would be the received size and the settings limit, both as `int`s.  Down the road, we can check our analytics and see how often people are banging into these limits and if they should be adjusted.

felliott:
LGTM!  Merged as 887b954894f557846c1156d271716b869a861707

PR #239, opened 2017-02-09
[SVCS-235] MFR handle 'field too large' error more gracefully

Johnetordoff:
# Purpose 

Our csv tabular render chokes when a field is too large to be read by the underlying library: https://sentry.cos.io/share/issue/32312e36363838/ This makes ot show a  more informative error message.

# Changes

Catches error thrown by _csv library and display field limit error or other error caught by _csv

# Side Effect

None that I know of

# Ticket

https://openscience.atlassian.net/browse/SVCS-235

# Figure
New Message
screen shot 2017-02-09 at 3 44 29 pm

### DIFF
--- a/mfr/extensions/tabular/render.py
+++ b/mfr/extensions/tabular/render.py
@@ -1,5 +1,6 @@
 import os
 import json
+import _csv
 
 from mako.lookup import TemplateLookup
 from mfr.core import extension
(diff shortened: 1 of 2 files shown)


mfr/extensions/tabular/render.py
```diff
@@ -1,5 +1,6 @@
 import os
 import json
+import _csv
```
felliott: `_csv` is the internal C module that python's `csv` module wraps.  `csv` re-exports `_csv.Error` as its own `Error`. Use that instead, since that should have a stable interface.

mfr/extensions/tabular/render.py
```diff
@@ -41,7 +42,14 @@ def _render_grid(self, fp, ext, *args, **kwargs):  # assets_path, ext):
         """
         self._renderer_tabular_metrics = {}
 
-        sheets = self._populate_data(fp, ext)
+        try:
+            sheets = self._populate_data(fp, ext)
+        except _csv.Error as e:
```
felliott: Is there a reason to handle this here instead of in the csv parsing library (`./libs/stdlib_tools.py`)?

mfr/extensions/tabular/render.py
```diff
@@ -41,7 +42,14 @@ def _render_grid(self, fp, ext, *args, **kwargs):  # assets_path, ext):
...
+            if any("field larger than field limit" in errorMsg for errorMsg in e.args):
+                raise exceptions.TabularRendererError('This file contains a field too large to render. '
```
felliott: `TabularRendererError` throws a 500 by default (well, one of its parents does), but I think a 40x would be more appropriate for this particular error.

felliott:
Merged as 3245d7ab0612f24920096df53cc9d7b4c6de93fc

PR #241, opened 2017-02-16
[hotfix][SVCS-257] Fix  MFR failing to render .ipynb files on prod

Johnetordoff:
# Purpose

Fix error where .ipynb files don't render using the MFR on production, but work locally.

# Changes

Change requirements.txt so IPython loads as part of main requirements.txt not  dev-requirements.txt 

# Side Effects

None that I know of.

# Ticket

https://openscience.atlassian.net/browse/SVCS-257

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,7 +24,7 @@ pydocx==0.7.0
 Pillow==2.8.2
 
 # IPython
-ipdb==0.10.2
+ipython==5.2.2
 nbconvert==4.2.0
 nbformat==4.1.0
 traitlets==4.2.2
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -24,6 +24,7 @@ pydocx==0.7.0
 Pillow==2.8.2
 
 # IPython
+ipdb==0.10.2
```
felliott: Why ipdb and not ipython?

Johnetordoff: I'm just trying to preserve all the dependencies we would normally use, it would be disappointing if we get it into prod and there's related module we don't know about which isn't present. But see why you we might want to be more direct.

felliott:
Merged as 8c11d79db2087e5556f3bc946ab791a8624c9d6e.  Thanks, @Johnetordoff!

PR #242, opened 2017-02-16
[SVCS-258] Allow confident use of 'inv' for MFR

Johnetordoff:
# Purpose

Hashing of aliases can cause the command `invoke` and it's alias `inv` to be from different versions of python. This causes a well know error that should be documented.

# Changes 

Add line to docs explaining error

# Side Effects

None that I know of.

# Tickets

https://openscience.atlassian.net/browse/SVCS-258


felliott:
Merged as f31504473692bcd40d66bcc55a18f13dc544e17d

PR #244, opened 2017-03-03
[SVCS-248] Add zip rendering

Johnetordoff:
# Purpose

This allows the MFR to render .zip files to an extent by listing there contents and absolute directories.

# Changes

These changes add a renderer and and .mako template for displaying the listed contents of a zip file.

# Side Effects

Sometimes the content rendered by the MFR is metadata typically hidden to a user using a zip file. This might be confusing. 
Example:
screen shot 2017-03-03 at 9 39 51 am
_this zip consists only of .txt files called test 1 and test 2_

Also it was difficult writing good tests for this renderer

# Ticket

https://openscience.atlassian.net/browse/SVCS-248


mfr/extensions/zip/templates/viewer.mako
```diff
@@ -0,0 +1,14 @@
+<link href="https://fonts.googleapis.com/css?family=Open+Sans:400,600,300,700" rel="stylesheet" type="text/css">
+<link rel="stylesheet" href="static/css/default.css">
+<div style="word-wrap: break-word;" class="mfrViewer">
+<code>
```
felliott: This tag is unclosed.

mfr/extensions/zip/render.py
```diff
@@ -0,0 +1,30 @@
...
+            os.path.join(os.path.dirname(__file__), 'templates')
+        ]).get_template('viewer.mako')
+
+    def render(self):
+        zip_file = zipfile.ZipFile(self.file_path, 'r')
+
+        return self.TEMPLATE.render(zipped_filenames=self.format_zip(zip_file))
+
+    def format_zip(self, zip_file):
+        return '<br>'.join(zip_file.namelist())
```
felliott: I'm having a hard time coming up with an actionable example, but since users decide the file name, please markup-escape these strings before rendering.  Bad example: if a user calls their file `<b>foo</b>`  the tags should not be interpreted as markup. Sample usage: https://github.com/CenterForOpenScience/modular-file-renderer/blob/81131736cdd44d27a5fece6d62bac1f52f017abe/mfr/core/provider.py#L23

mfr/extensions/zip/render.py
```diff
@@ -0,0 +1,31 @@
...
+            os.path.join(os.path.dirname(__file__), 'templates')
+        ]).get_template('viewer.mako')
+
+    def render(self):
+        zip_file = zipfile.ZipFile(self.file_path, 'r')
+
+        return self.TEMPLATE.render(zipped_filenames=self.format_zip(zip_file))
+
+    def format_zip(self, zip_file):
+        return '<br>'.join([markupsafe.escape(x) for x in zip_file.namelist()])
```
cslzchen: Both `namelist()` and `infolist()` contain duplicated entries. For example, `mfr-test.txt` has three entries where `__MACOSX/` and `__MACOSX/.__mfr-test.txt` are not what we want.  In addition, unlike `file`, `zip` and `dir` do not have the prefix. Is there a general rule that we know?

What's in my zip:

```
mfr-test/
mfr-test.zip
mfr-test.docx
mfr-test.pdf
mfr-test.pptx
mfr-test.txt
_mfr-test.txt
```
What shows up in OSF:

```
_mfr-test.txt
__MACOSX/
__MACOSX/.__mfr-test.txt
mfr-test/
mfr-test.docx
__MACOSX/._mfr-test.docx
mfr-test.pdf
__MACOSX/._mfr-test.pdf
mfr-test.pptx
__MACOSX/._mfr-test.pptx
mfr-test.txt
__MACOSX/._mfr-test.txt
mfr-test.zip
```

cslzchen: Another concern is that the prefix seems OS dependent. If we want to identify it, should we consider taking other OS into account?

tests/extensions/zip/test_renderer.py
```diff
@@ -0,0 +1,51 @@
...
+@pytest.fixture
+def renderer(metadata, test_file_path, url, assets_url, export_url):
+    return ZipRenderer(metadata, test_file_path, url, assets_url, export_url)
+
+
+class TestZipRenderer:
+
+    def test_render(self, renderer):
+        body = renderer.render()
+        assert 'test 1<br>__MACOSX/<br>__MACOSX/._test 1<br>test 2<br>__MACOSX/._test 2' in body
```
cslzchen: 1. As mentioned, `_MACOSX` is OS independent. It works here (and on travis) because the `zip` file we uploaded is compressed by OSX.
2. Maybe considering a more complicated zip file (directories, different file types, huge amount of files, etc.)

mfr/extensions/zip/templates/viewer.mako
```diff
@@ -0,0 +1,18 @@
...
+<div style="word-wrap: break-word;" class="mfrViewer">
+    <code>
+        <h1>
+            Zip File:
+        </h1>
+        <p>
+            Download the .zip file to view the contents.
+        <p>
+        <p>
+            ${zipped_filenames}
```
cslzchen: My thoughts continued on @felliott comment about `infolist()` (not in the scope of this ticket), we can improve the UI (a friendly list instead of a raw text output) and add some meta data about them.

tests/extensions/zip/test_renderer.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def test_render(self, renderer):
+        body = renderer.render()
+
+        assert '<td>test 2' in body and '<td>test 1' in body
+
+    def test_format_zip(self, renderer, zip_file):
+        zip_string = renderer.format_zip(zip_file)
+
+        assert '<td>test 2' in zip_string and '<td>test 1' in zip_string
```
felliott: Please expand your testing to match the expanded functionality of the zip renderer. Things I'd like to see:

* the timestamps are present and correct
* the byte size is present and correct
* proper formatting of larger byte sizes 1k, 1M, etc.  (for 1M, a megabyte text file of a single character should compress pretty well)
* `__MACOSX` files are actually being excluded
* What happens when you pass in something that isn't a zipfile?  Is the error message informative or vague?
* Is it possible to have an empty zipfile?  If so, we should state that instead of displaying an empty table.
* zipped zipfiles should *not* be recursively expanded.

mfr/extensions/zip/templates/viewer.mako
```diff
@@ -0,0 +1,19 @@
...
+<link rel="stylesheet" href="static/css/bootstrap.min.css">
+<link rel="stylesheet" href="static/css/default.css">
+<div style="word-wrap: break-word;" class="mfrViewer">
+    <code>
+        <h1>
+            Zip File:
+        </h1>
+        <p>
+            Download the .zip file to view the contents.
+        <p>
```
felliott: Shouldn't this be a `</p>`?

mfr/extensions/zip/render.py
```diff
@@ -0,0 +1,47 @@
+import os
+import zipfile
+import markupsafe
```
felliott: PEP8: `markupsafe` isn't in the stdlib.

requirements.txt
```diff
@@ -5,6 +5,7 @@ humanfriendly==2.1
 invoke==0.13.0
 mako==1.0.1
 raven==5.27.0
+requests==2.13.0
```
felliott: Why was this added?  It doesn't appear to be related.

requirements.txt
```diff
@@ -17,7 +18,6 @@ agent==0.1.2
 Pygments==2.0.2
 
 # Docx
-#git+https://github.com/jsocol/bleach.git
```
felliott: Why was this removed?  It doesn't appear to be related.

tests/extensions/zip/test_renderer.py
```diff
@@ -0,0 +1,61 @@
+import os
+import pytest
+from zipfile import ZipFile
+
+from mfr.core.provider import ProviderMetadata
+
+from mfr.extensions.zip import ZipRenderer
```
felliott: PEP8 the imports, please.

tests/extensions/zip/test_renderer.py
```diff
@@ -0,0 +1,61 @@
...
+
+BASE = os.path.dirname(os.path.abspath(__file__))
+
+
+@pytest.fixture
+def metadata():
+    return ProviderMetadata('test', '.zip', 'text/plain', '1234', 'http://wb.osf.io/file/test.zip?token=1234')
```
felliott: Long line

tests/extensions/zip/test_renderer.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@pytest.fixture
+def test_file_path():
+    return os.path.join(os.path.dirname(os.path.abspath(__file__)), 'files', 'test.zip')
+
+@pytest.fixture
+def test_render_body():
+    with open(os.path.join(os.path.dirname(os.path.abspath(__file__)), 'files', 'test_render_body.html'), 'r') as fp:
+        return fp.read()
```
felliott: This fixture doesn't appear to be used.

tests/extensions/zip/test_renderer.py
```diff
@@ -0,0 +1,61 @@
...
+    return ProviderMetadata('test', '.zip', 'text/plain', '1234', 'http://wb.osf.io/file/test.zip?token=1234')
+
+@pytest.fixture
+def zip_file():
+    return ZipFile(os.path.join(BASE, 'files', 'test.zip'), 'r')
+
+
+@pytest.fixture
+def test_file_path():
+    return os.path.join(os.path.dirname(os.path.abspath(__file__)), 'files', 'test.zip')
```
felliott: You could shorten this by using the `BASE` variable you declared.

mfr/extensions/zip/render.py
```diff
@@ -0,0 +1,53 @@
...
+        ]).get_template('viewer.mako')
+
+    def render(self):
+        zip_file = zipfile.ZipFile(self.file_path, 'r')
+
+        return self.TEMPLATE.render(zipped_filenames=self.format_zip(zip_file))
+
+    def format_zip(self, zip_file):
+
+        if type(zip_file) != zipfile.ZipFile:
```
felliott: Minor: `isinstance` is more concise.

mfr/extensions/zip/render.py
```diff
@@ -0,0 +1,53 @@
...
+                  '</thead>'\
+                  % ('File Name', 'Modified    ', 'Size')
+
+        for zinfo in filelist:
+            date = "%d-%02d-%02d %02d:%02d:%02d" % zinfo.date_time[:6]
+            message += "<tr><td>%-46s</td> <td>%s</td> <td>%s<td></tr>" %\
+                       (markupsafe.escape(zinfo.filename), date, sizeof_fmt(int(zinfo.file_size)))
+
+        message += '</table>'
+        return message
```
felliott: I don't know how I missed this before, but this method (`format_zip`) is essentially doing templating in code.  Why do that when we are already using a templating system (mako)?  You could move the HTML into the template.mako, and have `format_zip` build the structure to be fed to it.

mfr/extensions/zip/render.py
```diff
@@ -0,0 +1,53 @@
...
+            return 'This zip file is empty.'
+
+        message = '<table class="table table-hover">' \
+                  '<thead>' \
+                    '<th>%-46s</th><th>%19s</th><th>%12s</th>' \
+                  '</thead>'\
+                  % ('File Name', 'Modified    ', 'Size')
+
+        for zinfo in filelist:
+            date = "%d-%02d-%02d %02d:%02d:%02d" % zinfo.date_time[:6]
```
felliott: I think we might have discussed this already, but can we get any time zone information on this object?  Right now it doesn't display any time zone, which makes the timestamps much less informative.

mfr/extensions/zip/render.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def format_zip(self, zip_file):
+
+        if type(zip_file) != zipfile.ZipFile:
+            return 'This is not a valid zip file.'
+
+        filelist = [file for file in zip_file.filelist if not file.filename.startswith('__MACOSX')]
+
+        if not filelist:
+            return 'This zip file is empty.'
```
felliott: Please raise errors instead of returning a string.  E.g.: https://github.com/CenterForOpenScience/modular-file-renderer/blob/develop/mfr/extensions/jasp/render.py#L107

mfr/extensions/zip/templates/viewer.mako
```diff
@@ -0,0 +1,19 @@
+<link href="https://fonts.googleapis.com/css?family=Open+Sans:400,600,300,700" rel="stylesheet" type="text/css">
+<link rel="stylesheet" href="static/css/bootstrap.min.css">
+<link rel="stylesheet" href="static/css/default.css">
+<div style="word-wrap: break-word;" class="mfrViewer">
+    <code>
```
felliott: `<code>` is a phrasing element like `<strong>` or `<em>` and shouldn't contain block-level elements, like `<h1>` and `<p>`.  In my local env, this causes weird floating elements before and after the code.  If you want the content to be monospaced and have a background, you should do that with CSS.

felliott: Example:

screen shot 2017-11-07 at 12 15 13 pm

mfr/extensions/zip/render.py
Johnetordoff: This is not really the same use case, there's no error it's just an actually empty zip which is a valid though rare possiblity.

mfr/extensions/zip/render.py
Johnetordoff: No time zone data is available.

felliott:
Looks good, moving to RTM!  Thanks, @Johnetordoff!

PR #245, opened 2017-03-08
[SVCS-261] Fix svg rendering

Johnetordoff:
# Purpose

fix svg rendering currently broken on production

# Changes

Uses Svg render to load the file directly into the html.

# Side Effect

None that I know of.

# Ticket

https://openscience.atlassian.net/browse/SVCS-261


felliott:
Declining this for now.  Displaying other people's arbitrary SVG is a security risk.  We'll either need to figure out a way to sanitize it, or else stick to converting it to pdf outside the browser.

PR #248, opened 2017-03-21
[SVCS-140] Fix to allow >= 31.0 setuptools

Johnetordoff:
## Purpose

Let MFR run later versions of setuptools (>= 31.0 setuptools)

## Changes

Creates version.py file to prevent setuptools to prevent circular import.

## Side Effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/SVCS-140

## WB PR
https://github.com/CenterForOpenScience/waterbutler/pull/201


version.py
```diff
@@ -0,0 +1 @@
+__version__ = '0.20.1'
```
cslzchen: @felliott this will be the file that needs manual version bump for release/hotfix
@Johnetordoff need to update this to 0.22.0

cslzchen:
And a follow-up: we should update the version of `setuptools` to their latest 37.0.0, which works locally for me.

cslzchen:
In addition, there are a few other places, for example `remote_logging.py`, `conf.py`, etc. Please take a thorough look into the code base.

cslzchen:
@Johnetordoff this continues in a new PR: https://github.com/CenterForOpenScience/modular-file-renderer/pull/301 where I fixed a few minor issues and rebase all the commits into one.

Issue #252, opened 2017-04-28
Commands sent to docker daemon hang when starting a Libre Office server inside the MFR container


rija:
I'm running MFR in a Docker container using the supplied Dockerfile,
When I starts  Libre Office Server in the container using ``unoconv --listener`` to allow preview of Office documents, the docker container  becomes irresponsive to control (cannot exit the shell, docker stop, restart or exec hang ).

I have to shutdown the docker-machine and restart in order to gain access of the container again.

I run docker on  Mac OS X, using docker-machine with Docker v1.12.0

below is how I start my container

```
$ git clone http://github.com/centerforopenscience/modular-file-renderer/
$ cd modular-file-renderer

$ docker build -t="mfr" .

$ docker run -d -p 7778:7778 \
   -e SERVER_CONFIG_ADDRESS=0.0.0.0 \
   -e SERVER_CONFIG_PROVIDER_NAME=http \
   -e SERVER_CONFIG_ALLOWED_PROVIDER_DOMAINS=http://www.example.com/  \
   --name mfr_server \
   mfr
```

and how I run the Libre Office server:

```
 $ docker exec -it mfr bash
 $ unoconv --listener &
```
which will launch  __soffice.bin__ process

If I kill the __unoconv__ process, the __soffice.bin__ still run.

If I try to exit the shell, the Docker daemon hangs.
I have to ``kill -9`` the docker commands and restart the  docker-machine in order to regain control of the container.

## investigation

There may be a bug in unoconv (see https://github.com/dagwieers/unoconv/issues/189)  but irrespective of that , I think it's not a good idea to have mfr server running as PID 1: The lifecycle of other processes that we may want to run on the container won't be started or shutdown gracefully by an init process.

using a process supervisor as PID1 is a better approach.

Even if a MFR is intended to be the only process running, it's more portable to use the logging/lifecycle/process control features of a process manager than rely on docker for this.

## workaround

* immediate: before exiting the shell, kill the __soffice.bin__ manually
* proper: re-engineer the Dockerfile so a process manager (e.g: supervisord or mozilla circus) is used as PID 1

felliott:
Thanks, @rija!  A bit of background: the unoconv container has been troublesome for us, frequently locking up and very hard to upgrade.  @TomBaxter was working on trying to upgrade it for us, and I think he ran into this issue before.  Tom, do you mind taking a look at this PR and see if it fixes some of the issues we were encountering?

Cheers,
@felliott

[...]

PR #253, opened 2017-04-28
Updated Dockerfile so that PID 1 is attached to a process supervisor …

rija:
**A proposed fix for #252** 

I've chosen Mozilla Circus as process manager because:
* as of time of writing, supervisord don't seem to work with python 3.*
* I had hard time with installing/configuring supervisord on Centos/Redhat installs
* Circus can manage both processes and sockets which could help productionize an mfr install behind WSGI servers

**changes:**

Added configuration file for Circus to allow circus to manage invoke server and headless Libre Office server processes cleanly
Copied MFR config file to mfr-test.json and waterbutler-test.json in the docker container to silence warnings and allow for in-container configuration
Moved the CMD and EXPOSE to the top of the Dockerfile to take advantage of the Docker build caching mechanism for iterative builds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,23 @@
 FROM python:3.5-slim
 
+#
+# * example usage:
+#
+# $ docker build -t="mfr" .
+#
+# $ docker run -d -p 7778:7778 \
+#    -e SERVER_CONFIG_ADDRESS=0.0.0.0 \
+#    -e SERVER_CONFIG_PROVIDER_NAME=http \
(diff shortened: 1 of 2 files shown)


felliott:
Hey @rija,

We prefer to do one-process-per-container, and we've added a unoconv service to the OSF's docker-compose file (see my response to your issue #252).  I don't think we want to adopt a process supervisor at this time, so I'm going to close this.  Thank you!

Cheers,
@felliott

PR #255, opened 2017-05-02
[SVCS-76] Track requests within OSF

Johnetordoff:
# Purpose

Dropbox and Github have a nice feature where they return request identifiers in their response headers. That way, when a request starts acting up you can contact their support with the ID. This enables MFR to do the same thing. Now we can verify a particular request made from the browser is the same one in our logs and more easily track inter-service requests if we want.

# Changes

Simply adds a uniquely generated id to the header "X-MFR-REQUEST-ID" when the request is being prepared.

# Side Effects
None that I know of.

# Ticket

https://openscience.atlassian.net/browse/SVCS-76

### DIFF
--- a/mfr/server/handlers/core.py
+++ b/mfr/server/handlers/core.py
@@ -132,6 +132,7 @@ class BaseHandler(CorsMixin, tornado.web.RequestHandler, SentryMixin):
         )
 
         self.source_file_id = uuid.uuid4()
+        self.add_header('X-MFR-REQUEST-ID', str(uuid.uuid4()))
 
     async def write_stream(self, stream):
         try:


mfr/server/handlers/core.py
```diff
@@ -132,6 +132,7 @@ def NAME(self):
         )
 
         self.source_file_id = uuid.uuid4()
+        self.add_header('X-MFR-REQUEST-ID', str(uuid.uuid4()))
```
cslzchen: I have a few further questions on how this work.

1. Do servers (WB and MFR) need to handle this customized header explicitly? For example, is it automatically logged?

2. Why not let client generate the header? (WB and MFR can generate it only when client does not). We cannot guarantee that the client always receives a response that contains this header, especially when exceptions occur.

3. For "**It would be doubly nice if somehow we could propagate this request id when a request involves multiple services.**": can we pass this ID along the requests pipeline (including requests to different OSF services which belong to the same request cycle?

4. `X-REQUEST-ID` is a the general name. What's the pros and cons of using `X-MFR-REQUEST-ID` and `X-WB-REQUEST-ID`?

Johnetordoff: My understanding is that the immediate use for this is that it is visible in the sentry logs, so if a user can give us their request id we can easily associate that with the logs, other then that other uses are speculative.

felliott:
I'm good with this as a first step.  I'll make new tickets to expand the functionality (request id propagation, etc.)

PR #256, opened 2017-05-05
Catch KEEN_PRIVATE_PROJECT_ID is not defined

sonnenschein123:
In the default configuration is the value KEEN_PRIVATE_PROJECT_ID commented out, so if you run the code you get the exception "AttributeError: module 'mfr.server.settings' has no attribute 'KEEN_PRIVATE_PROJECT_ID'". 
This catches the exception.

### DIFF
--- a/mfr/core/remote_logging.py
+++ b/mfr/core/remote_logging.py
@@ -17,7 +17,10 @@ logger = logging.getLogger(__name__)
 
 async def log_analytics(request, metrics, is_error=False):
     """Send events to Keen describing the action that occurred."""
-    if settings.KEEN_PRIVATE_PROJECT_ID is None:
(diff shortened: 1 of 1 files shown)


felliott:
Hey @sonnenschein123!

Where is `KEEN_PRIVATE_PROJECT_ID` commented out?  It looks like it's initialized to `None` [here](https://github.com/CenterForOpenScience/modular-file-renderer/blob/0329595504508314956cda8384e39d230a4d05c1/mfr/server/settings.py#L48).  Is there somewhere in the docs that we suggest that should be commented out?

Cheers,
@felliott

sonnenschein123:
hey @felliott 
thanks

PR #258, opened 2017-05-15
[SVCS-129] Don't 500 when missing required query parameters

Johnetordoff:
# Purpose 

Making an MFR export request without passing the format parameter causes a 500. It
should throw a 400 instead. There are probably other unhandled cases like this. Check for missing required query params and throw meaningful errors.

Effected Params are:
url
format

# Changes

Adds exception handling and unit tests to MFR.

# Ticket
https://openscience.atlassian.net/browse/SVCS-129

### DIFF
--- a/mfr/server/handlers/export.py
+++ b/mfr/server/handlers/export.py
@@ -25,7 +25,8 @@ class ExportHandler(core.BaseHandler):
 
         format = self.request.query_arguments.get('format', None)
         if not format:
-            raise InvalidParameters("Invalid Request: <your error message>")
+            raise InvalidParameters("Invalid Request: Url requires query parameter 'format' with"
+                                    " appropriate extension")
         # TODO: do we need to catch exceptions for decoding?
         self.format = format[0].decode('utf-8')
 
--- a/mfr/server/handlers/export.py
+++ b/mfr/server/handlers/export.py
@@ -3,7 +3,7 @@ import asyncio
 import logging
 
 import waterbutler.core.streams
-import waterbutler.core.exceptions
+from waterbutler.core.exceptions import InvalidParameters, DownloadError
 
 from mfr.server import settings
 from mfr.core import utils as utils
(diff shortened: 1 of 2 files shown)


tests/server/handlers/test_query_params.py
```diff
@@ -0,0 +1,28 @@
...
+
+class TestRenderHandler(utils.HandlerTestCase):
+
+    @testing.gen_test
+    def test_format_url(self):
+
+        with pytest.raises(HTTPError) as e:
+            yield self.http_client.fetch(self.get_url('/export'), method='GET')
+        assert e.value.message == 'Bad Request'
+        assert e.value.code == 400
```
cslzchen: Consider using `http.HTTPStatus` ([link for 3.5](https://docs.python.org/3.5/library/http.html#http.HTTPStatus)).

```python
from http import HTTPStatus
...
    assert e.value.code == HTTPStatus.BAD_REQUEST
...
```

mfr/server/handlers/export.py
```diff
@@ -24,7 +24,11 @@ class ExportHandler(core.BaseHandler):
 
         await super().prepare()
 
-        self.format = self.request.query_arguments['format'][0].decode('utf-8')
+        if self.request.query_arguments.get('format'):
```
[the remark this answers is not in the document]
cslzchen: One more comment: are there other query parameters that should be handled this way as well?

tests/server/handlers/test_query_params.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+from http import HTTPStatus
+
+from tornado.httpclient import HTTPError
+from tests import utils
+from tornado import testing
```
cslzchen: `pytest` is not a standard python library.

```python
from http import HTTPStatus

import pytest
from tornado import testing
from tornado.httpclient import HTTPError

from tests import utils

```

mfr/server/handlers/export.py
```diff
@@ -24,7 +23,12 @@ class ExportHandler(core.BaseHandler):
 
         await super().prepare()
 
-        self.format = self.request.query_arguments['format'][0].decode('utf-8')
+        format = self.request.query_arguments.get('format', None)
+        if not format:
+            raise InvalidParameters("Invalid Request: <your error message>")
```
felliott: ...

tests/server/handlers/test_query_params.py
felliott: While you're fixing up the error message, go ahead and do this, too.

mfr/server/handlers/export.py
```diff
@@ -1,15 +1,14 @@
...
+import os
 
+from waterbutler.core.exceptions import InvalidParameters, DownloadError
 import waterbutler.core.streams
-import waterbutler.core.exceptions
 
+from mfr.core import utils
 from mfr.server import settings
-from mfr.core import utils as utils
 from mfr.server.handlers import core
```
felliott: :+1: for all of this

PR #259, opened 2017-05-19
[RFR][SVCS-336]MFR issue: cannot render (videos, PDFs, or PDBs) locally when dockerized

TomBaxter:
## Purpose:
In local development, those MFR extensions that create javascript links to WB files, must use localhost instead of 192.168.168.167. This is because the request will come from the browser instead of the MFR docker instance. 

Add catch for rendering in local development

## Changes:
Add EXTENSION settings for LOCAL_HOST and DOCKER_LOCAL_HOST
Add mfr.extension.utils.get_local_export_url to ...
    1. Switch DOCKER_LOCAL_HOST to LOCAL_HOST
    2. Remove "mode" parameter
For renderers pdb, pdf, and video:
    Check for DOCKER_LOCAL_HOST and call mfr.extension.utils.get_local_export_url

## Side effects
None

## Important
There is a [related PR](https://github.com/CenterForOpenScience/osf.io/pull/7740) That adds EXTENSION_CONFIG_LOCAL_DEVELOPMENT=1 to .docker-compose.mfr.env . Would be nice if the were merged together, though nothing bad would happen if they weren't. 

[#SVCS-336]

- [x] Create PR for OSF.io add this line to OSF.io .docker-compose.mfr.env ```EXTENSION_CONFIG_LOCAL_DEVELOPMENT=1```

### DIFF
--- a/mfr/extensions/pdb/render.py
+++ b/mfr/extensions/pdb/render.py
@@ -5,6 +5,7 @@ import json
 from mako.lookup import TemplateLookup
 
 from mfr.core import extension
+from mfr.extensions.utils import download_from_template
 from mfr.extensions.pdb import settings
 
 
@@ -15,10 +16,11 @@ class PdbRenderer(extension.BaseRenderer):
             os.path.join(os.path.dirname(__file__), 'templates')
         ]).get_template('viewer.mako')
 
+    @download_from_template
     def render(self):
         return self.TEMPLATE.render(
             base=self.assets_url,
-            url=self.metadata.download_url,
+            url=self.download_url.geturl(),
             options=json.dumps(settings.OPTIONS),
         )
 
--- a/mfr/extensions/pdb/render.py
+++ b/mfr/extensions/pdb/render.py
@@ -5,8 +5,8 @@ import json
 from mako.lookup import TemplateLookup
 
 from mfr.core import extension
-from mfr.extensions.utils import download_from_template
 from mfr.extensions.pdb import settings
+from mfr.extensions.utils import munge_url_for_localdev
 
 
 class PdbRenderer(extension.BaseRenderer):
@@ -16,7 +16,7 @@ class PdbRenderer(extension.BaseRenderer):
             os.path.join(os.path.dirname(__file__), 'templates')
         ]).get_template('viewer.mako')
 
-    @download_from_template
+    @munge_url_for_localdev
     def render(self):
         return self.TEMPLATE.render(
             base=self.assets_url,
--- a/mfr/extensions/pdf/render.py
+++ b/mfr/extensions/pdf/render.py
@@ -3,7 +3,7 @@ import os
 from mako.lookup import TemplateLookup
 
 from mfr.core import extension
-from mfr.extensions.utils import download_from_template
+from mfr.extensions.utils import munge_url_for_localdev
 
 
 class PdfRenderer(extension.BaseRenderer):
@@ -13,7 +13,7 @@ class PdfRenderer(extension.BaseRenderer):
             os.path.join(os.path.dirname(__file__), 'templates')
         ]).get_template('viewer.mako')
 
-    @download_from_template
+    @munge_url_for_localdev
     def render(self):
         return self.TEMPLATE.render(base=self.assets_url, url=self.download_url.geturl())
 
--- a/mfr/extensions/pdf/render.py
+++ b/mfr/extensions/pdf/render.py
@@ -3,6 +3,7 @@ import os
 from mako.lookup import TemplateLookup
 
 from mfr.core import extension
+from mfr.extensions.utils import download_from_template
 
 
 class PdfRenderer(extension.BaseRenderer):
@@ -12,8 +13,9 @@ class PdfRenderer(extension.BaseRenderer):
             os.path.join(os.path.dirname(__file__), 'templates')
         ]).get_template('viewer.mako')
 
+    @download_from_template
     def render(self):
-        return self.TEMPLATE.render(base=self.assets_url, url=self.metadata.download_url)
+        return self.TEMPLATE.render(base=self.assets_url, url=self.download_url.geturl())
 
     @property
     def file_required(self):
--- a/mfr/extensions/settings.py
+++ b/mfr/extensions/settings.py
@@ -1,0 +1,11 @@
+from mfr import settings
+
+config = settings.child('EXTENSION_CONFIG')
+
+# LOCAL_DEVELOPMENT must be set to true for extensions that produce a waterbutler URL in the
+# TEMPLATE to render, when using local docker development environment. Call will be coming from
+# browser instead of docker container.
+
+LOCAL_DEVELOPMENT = config.get_bool('LOCAL_DEVELOPMENT', 0)
+DOCKER_LOCAL_HOST = config.get('DOCKER_LOCAL_HOST', '192.168.168.167')
+LOCAL_HOST = config.get('LOCAL_HOST', 'localhost')
--- a/mfr/extensions/utils.py
+++ b/mfr/extensions/utils.py
@@ -1,0 +1,22 @@
+from urllib.parse import urlparse, parse_qs, urlencode
+from mfr.extensions import settings
+
+
+def download_from_template(func):
+    """
+    Check if LOCAL_DEVELOPMENT and if so
+    Replace the docker domain with localhost (for use in local development).
+    e.g. http://localhost:7777/foo/bar => http://192.168.168.167:7777/foo/bar
+    """
+    def wrapper(self):
+        download_url = urlparse(self.metadata.download_url)
+        if (settings.LOCAL_DEVELOPMENT and download_url.hostname == settings.DOCKER_LOCAL_HOST):
+            query_dict = parse_qs(download_url.query, keep_blank_values=True)
+            query_dict.pop('mode', None)
+            download_url = download_url._replace(query=urlencode(query_dict, doseq=True))
+            download_url = download_url._replace(
+                netloc="{}:{}".format(settings.LOCAL_HOST, download_url.port)
+            )
+        self.download_url = download_url
+        return func(self)
+    return wrapper
--- a/mfr/extensions/utils.py
+++ b/mfr/extensions/utils.py
@@ -1,21 +1,22 @@
 from urllib.parse import urlparse, parse_qs, urlencode
+
 from mfr.extensions import settings
 
 
-def download_from_template(func):
+def munge_url_for_localdev(func):
     """
-    Check if LOCAL_DEVELOPMENT and if so
-    Replace the docker domain with localhost (for use in local development).
-    e.g. http://localhost:7777/foo/bar => http://192.168.168.167:7777/foo/bar
+    If MFR is being run in a local development environment (i.e. LOCAL_DEVELOPMENT is True), we
+    need to replace the internal host (the one the backend services communicate on, default:
+    192.168.168.167) with the external host (the one the user provides, default: "localhost")
+    e.g. http://192.168.168.167:7777/foo/bar => http://localhost:7777/foo/bar
     """
     def wrapper(self):
         download_url = urlparse(self.metadata.download_url)
         if (settings.LOCAL_DEVELOPMENT and download_url.hostname == settings.DOCKER_LOCAL_HOST):
             query_dict = parse_qs(download_url.query, keep_blank_values=True)
-            query_dict.pop('mode', None)
-            download_url = download_url._replace(query=urlencode(query_dict, doseq=True))
             download_url = download_url._replace(
-                netloc="{}:{}".format(settings.LOCAL_HOST, download_url.port)
+                query=urlencode(query_dict, doseq=True),
+                netloc='{}:{}'.format(settings.LOCAL_HOST, download_url.port)
             )
         self.download_url = download_url
         return func(self)
--- a/mfr/extensions/video/render.py
+++ b/mfr/extensions/video/render.py
@@ -2,6 +2,7 @@ import os
 
 from mako.lookup import TemplateLookup
 
+from mfr.extensions.utils import download_from_template
 from mfr.core import extension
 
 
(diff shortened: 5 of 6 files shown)


mfr/extensions/video/render.py
```diff
@@ -13,7 +14,14 @@ class VideoRenderer(extension.BaseRenderer):
         ]).get_template('viewer.mako')
 
     def render(self):
-        return self.TEMPLATE.render(base=self.assets_url, url=self.url)
+        self_url = urlparse(self.url)
```
abought: This problem could more generally also affect other renderers that directly serve the file content, such as PDB files. Might be worth checking for similar edge case bugs?

TomBaxter: Thanks. Yep, already looking at updating pdf renderer.

abought: Wait, so it also affects a third filetype? Good find. Hope we can DRY it a bit.

mfr/extensions/video/render.py
```diff
@@ -13,7 +15,11 @@ class VideoRenderer(extension.BaseRenderer):
         ]).get_template('viewer.mako')
 
     def render(self):
-        return self.TEMPLATE.render(base=self.assets_url, url=self.url)
+        self_url = urlparse(self.metadata.download_url)
+        # Catch for local developement.
```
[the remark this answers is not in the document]
TomBaxter: complete

mfr/extensions/settings.py
```diff
@@ -0,0 +1,6 @@
+from mfr import settings
+
+config = settings.child('EXTENTION_CONFIG')
```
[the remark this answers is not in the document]
TomBaxter: complete

mfr/extensions/utils.py
```diff
@@ -0,0 +1,11 @@
+from urllib.parse import parse_qs, urlencode
+
+from mfr.extensions import settings
+
+def development_export_url(urlparsed_url):
+        query_dict = parse_qs(urlparsed_url.query, keep_blank_values=True)
+        query_dict.pop('mode', None)
```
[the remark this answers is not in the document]
TomBaxter: It's been quite a while since I worked on this and don't recall off the top of my head. In brief testing I was not able to come up with a good answer other then that it isn't needed. I can remove the pop, but it's very possible I had a good reason that I no longer recall.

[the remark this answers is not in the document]
cslzchen: This only affects local development. We can keep it there with the `TODO` note for you to recall the reason. :D

mfr/extensions/utils.py
```diff
@@ -0,0 +1,11 @@
...
+        urlparsed_url = urlparsed_url._replace(query=urlencode(query_dict, doseq=True))
+        urlparsed_url = urlparsed_url._replace(netloc="{}:{}".format(settings.LOCAL_HOST,
+                                                                     urlparsed_url.port))
+        return urlparsed_url
```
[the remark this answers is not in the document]
TomBaxter: Complete

mfr/extensions/pdb/render.py
```diff
@@ -16,9 +19,14 @@ class PdbRenderer(extension.BaseRenderer):
         ]).get_template('viewer.mako')
 
     def render(self):
+        self_url = urlparse(self.metadata.download_url)
+        # Catch for local developement.
+        if self_url.hostname == ext_settings.DOCKER_LOCAL_HOST:
```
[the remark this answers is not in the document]
TomBaxter: I see what you mean. However, I like the explicit DOCKER_LOCAL_HOST setting. I think it's less confusing. When someone sees the 192.168.168.167 address, they know just what we're talking about. Also, I like the explicitness of specifying the IP we will change, and no other. I'll split the difference with you and add another config LOCAL_DEVELOPMENT.

mfr/extensions/pdf/render.py
```diff
@@ -13,7 +15,11 @@ class PdfRenderer(extension.BaseRenderer):
         ]).get_template('viewer.mako')
 
     def render(self):
-        return self.TEMPLATE.render(base=self.assets_url, url=self.metadata.download_url)
+        self_url = urlparse(self.metadata.download_url)
+        # Catch for local developement.
+        if self_url.hostname == settings.DOCKER_LOCAL_HOST:
```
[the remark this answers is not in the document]
TomBaxter: Same as mentioned above.

mfr/extensions/pdb/render.py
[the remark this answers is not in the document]
cslzchen: Agreed. What you have looks good! 👍 
And probably no need for the `flag` I suggested. It leads to confusion and developers need to update local configuration.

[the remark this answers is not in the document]
TomBaxter: DOH! too late already did it, though I can undo it if you don't like it. You will now need to either add LOCAL_DEVELOPMENT=1 to your ~.cos file for MFR or...
add this line to your OSF.io .docker-compose.mfr.env...
```EXTENSION_CONFIG_LOCAL_DEVELOPMENT=1```
I'll make a coresponding OSF.io PR when this PR is finalized.

**Comment by TomBaxter** [comment]
RE: general question
I updated the PR Purpose: and code doc with...

In local development, those MFR extensions that create links to WB files in TEMPLATE, must use localhost instead of 192.168.168.167. This is because the request will come from the browser instead of the MFR docker instance.

File: mfr/extensions/pdb/render.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (cslzchen) [reply mfr/extensions/pdb/render.py]: I tried not to disturb the PR too much. It is better to have both in place (and it is) to make sure this only works for local docker environment.

**Comment by TomBaxter** [comment]
[related PR](https://github.com/CenterForOpenScience/osf.io/pull/7740) created.

File: mfr/extensions/pdb/render.py [hunk 8]
```diff
@@ -16,9 +19,14 @@ class PdbRenderer(extension.BaseRenderer):
         ]).get_template('viewer.mako')
 
     def render(self):
+        self_url = urlparse(self.metadata.download_url)
```
felliott: The code here and in the `render` methods for the pdf and video exporters are almost identical.  Could you move this into a decorator/utility method/base class/whatever to avoid the code duplication?

mfr/extensions/utils.py
```diff
@@ -0,0 +1,22 @@
+from urllib.parse import parse_qs, urlencode
+
+from mfr.extensions import settings
+
+
+def get_local_export_url(urlparsed_url):
+    """
+    Replace the docker domain with localhost (for use in local development).
+
```
felliott: Minor: show  a simple example, like:

````
http://localhost:7777/foo/bar => http://192.168.168.167:7777/foo/bar
````

File: mfr/extensions/utils.py [hunk 10]
```diff
@@ -0,0 +1,22 @@
...
+    :param parsed_url: the default export URL which is parsed by `urllib.parse.urlparse`
+    :return: the tweaked export URL for local development
+    """
+
+    query_dict = parse_qs(urlparsed_url.query, keep_blank_values=True)
+    # TODO: why do we pop `mode`?
```
felliott: ???  Did this come from somewhere else?

mfr/extensions/utils.py
```diff
@@ -0,0 +1,22 @@
+from urllib.parse import parse_qs, urlencode
+
+from mfr.extensions import settings
+
+
+def get_local_export_url(urlparsed_url):
+    """
+    Replace the docker domain with localhost (for use in local development).
+
+    :param parsed_url: the default export URL which is parsed by `urllib.parse.urlparse`
```
felliott: Your docstring is out-of-date.

mfr/extensions/utils.py
```diff
@@ -0,0 +1,22 @@
...
+    :return: the tweaked export URL for local development
+    """
+
+    query_dict = parse_qs(urlparsed_url.query, keep_blank_values=True)
+    # TODO: why do we pop `mode`?
+    query_dict.pop('mode', None)
+    urlparsed_url = urlparsed_url._replace(query=urlencode(query_dict, doseq=True))
+    urlparsed_url = urlparsed_url._replace(
+        netloc="{}:{}".format(settings.LOCAL_HOST, urlparsed_url.port)
+    )
```
felliott: I'm not familiar with the `_replace` method.  What does it do, and why are you using a private method?  Is there another way to do this?

mfr/extensions/utils.py
TomBaxter: completed.

mfr/extensions/utils.py
TomBaxter: Completed, I think?

mfr/extensions/utils.py
TomBaxter: That was a question from @cslzchen that I couldn't remember the answer to.

mfr/extensions/utils.py
TomBaxter: I'm sure this can be done many other ways, but I was trying to do the least amount of converting. This way I am just manipulating the ParseResult, a namedtuple subclass. Dropping the "mode" query parameter and switching the netloc.

https://docs.python.org/3.6/library/collections.html#collections.somenamedtuple._replace

mfr/extensions/pdb/render.py
TomBaxter: Completed.

mfr/extensions/pdb/render.py
```diff
@@ -5,6 +5,7 @@
 from mako.lookup import TemplateLookup
 
 from mfr.core import extension
+from mfr.extensions.utils import download_from_template
```
felliott: Nitpick: follow PEP8 import order, this should go after `mfr.extensions.pdb`

mfr/extensions/utils.py
```diff
@@ -0,0 +1,22 @@
+from urllib.parse import urlparse, parse_qs, urlencode
+from mfr.extensions import settings
```
felliott: nitpick: leave one space between import stanzas

mfr/extensions/settings.py
```diff
@@ -0,0 +1,11 @@
+from mfr import settings
+
+config = settings.child('EXTENSION_CONFIG')
+
+# LOCAL_DEVELOPMENT must be set to true for extensions that produce a waterbutler URL in the
+# TEMPLATE to render, when using local docker development environment. Call will be coming from
+# browser instead of docker container.
+
+LOCAL_DEVELOPMENT = config.get_bool('LOCAL_DEVELOPMENT', 0)
```
felliott: Nitpick: `0` ==> `False`

mfr/extensions/utils.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def download_from_template(func):
```
felliott: The decorator looks good to me, but I'd like to bikeshed the name, `download_from_template` is a little nonspecific.  Maybe `munge_url_for_localdev` or something like that?  Or even `munge_download_url` if this might end up doing more than just rewriting the localdev download url.  What do you think?

mfr/extensions/utils.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+    Check if LOCAL_DEVELOPMENT and if so
+    Replace the docker domain with localhost (for use in local development).
+    e.g. http://localhost:7777/foo/bar => http://192.168.168.167:7777/foo/bar
+    """
+    def wrapper(self):
+        download_url = urlparse(self.metadata.download_url)
+        if (settings.LOCAL_DEVELOPMENT and download_url.hostname == settings.DOCKER_LOCAL_HOST):
+            query_dict = parse_qs(download_url.query, keep_blank_values=True)
+            query_dict.pop('mode', None)
```
felliott: Please remove this and see what breaks.  If it's needed, add it back with a comment explaining why.  If not, leave it out.  If we can't figure out why we need it, we should leave it out until we know why.

mfr/extensions/utils.py
```diff
@@ -0,0 +1,22 @@
...
+            download_url = download_url._replace(query=urlencode(query_dict, doseq=True))
+            download_url = download_url._replace(
```
felliott: First, my bad: I thought the leading underscore meant this was a private method.  Clearly I need to learn more python.

Nitpick: you can modify both fields of the named tuple in a single `_replace` command.

mfr/extensions/utils.py
```diff
@@ -0,0 +1,22 @@
...
+                netloc="{}:{}".format(settings.LOCAL_HOST, download_url.port)
```
felliott: Super-nitpick: please use single quotes for literals.

mfr/extensions/video/render.py
```diff
@@ -2,6 +2,7 @@
 
 from mako.lookup import TemplateLookup
 
+from mfr.extensions.utils import download_from_template
 from mfr.core import extension
```
felliott: nitpick: import order

mfr/extensions/utils.py
```diff
@@ -0,0 +1,22 @@
+from urllib.parse import urlparse, parse_qs, urlencode
+from mfr.extensions import settings
+
+
+def download_from_template(func):
+    """
+    Check if LOCAL_DEVELOPMENT and if so
+    Replace the docker domain with localhost (for use in local development).
```
felliott: Copyedit:  "If MFR is being run in a local development environment (i.e. LOCAL_DEVELOPMENT is True), we need to replace the external host (the one the user provides, default: "localhost") with the internal host (the one the backend services communicate on, default: 192.168.168.167)."

mfr/extensions/utils.py
TomBaxter: Wasn't able to reproduce an issue, so removing. Completed.

mfr/extensions/settings.py
felliott: Everything else looks good, so I'll take care of it.  But generally the settings should follow the idiom they're defined in. This is a python config file, so `True` and `False` are appropriate.  The `.docker-compose.*.env` files are env files, so `1` and `0` are appropriate for them.

felliott:
A!W!E!S!O!M!E!  Moving to RTM.

PR #260, opened 2017-05-25
[SVCS-233] Fix KeyError: "Key not found: 'X-WATERBUTLER-METADATA'"

TomBaxter:
## Purpose:
Better error handling in mfr/providers/osf/provider.py

## Changes:
Catch error in metadata request earlier to avoid KeyError

## Side effects
None

[#SVCS-233]

### DIFF
--- a/mfr/providers/osf/provider.py
+++ b/mfr/providers/osf/provider.py
@@ -64,25 +64,26 @@ class OsfProvider(provider.BaseProvider):
             # TODO Remove this when API v0 is officially deprecated
             self.metrics.add('metadata.wb_api', 'v0')
             metadata_url = download_url.replace('/file?', '/data?', 1)
-            metadata_request = await self._make_request('GET', metadata_url)
-            metadata = await metadata_request.json()
+            metadata_response = await self._make_request('GET', metadata_url)
+            metadata = await metadata_response.json()
+            await metadata_response.release()
         else:
             # URL is for WaterButler v1 API
             self.metrics.add('metadata.wb_api', 'v1')
-            metadata_request = await self._make_request('HEAD', download_url)
-            # To make changes to current code as minimal as possible
-            try:
-                metadata = {'data': json.loads(metadata_request.headers['x-waterbutler-metadata'])['attributes']}
-                await metadata_request.release()
-            except KeyError:
-                raise exceptions.MetadataError(
-                    'Failed to fetch metadata. Received response '
-                    'code {}'.format(str(metadata_request.status)),
+            metadata_response = await self._make_request('HEAD', download_url)
+            response_code = metadata_response.status
+            response_reason = metadata_response.reason
+            response_headers = metadata_response.headers
+            await metadata_response.release()
+            if response_code != 200:
+                raise exceptions.MetadataError('Failed to fetch file metadata from WaterButler. Received response: code {} {}'.format(str(response_code), str(response_reason)),
                     metadata_url=download_url,
-                    response=await metadata_request.text(),
+                    response=response_reason,
                     provider=self.NAME,
-                    code=400,
-                )
+                    code=400)
+
+            try:
+                metadata = {'data': json.loads(response_headers['x-waterbutler-metadata'])['attributes']}
             except ContentEncodingError:
                 pass  # hack: aiohttp tries to unzip empty body when Content-Encoding is set
 


mfr/providers/osf/provider.py
```diff
@@ -66,23 +66,24 @@ def __init__(self, request, url):
...
-                    'code {}'.format(str(metadata_request.status)),
+            request_code = metadata_request.status
+            request_reason = metadata_request.reason
+            request_headers = metadata_request.headers
+            await metadata_request.release()
+            if request_code != 200:
+                raise exceptions.MetadataError('Failed to fetch file metadata from WaterButler. Received response: code {} {}'.format(str(request_code), str(request_reason)),
                     metadata_url=download_url,
-                    response=await metadata_request.text(),
+                    response=request_reason,
```
cslzchen: Does `response=` expect the response body instead of the status message? Status message is probably good enough but I just want to confirm this change. Here is the reference: `.reason` ([link](http://aiohttp.readthedocs.io/en/stable/client_reference.html#aiohttp.ClientResponse.reason)) and `.text()` ([link](http://aiohttp.readthedocs.io/en/stable/client_reference.html#aiohttp.ClientResponse.text)).

mfr/providers/osf/provider.py
```diff
@@ -66,23 +66,24 @@ def __init__(self, request, url):
...
-            except KeyError:
-                raise exceptions.MetadataError(
-                    'Failed to fetch metadata. Received response '
-                    'code {}'.format(str(metadata_request.status)),
+            request_code = metadata_request.status
+            request_reason = metadata_request.reason
+            request_headers = metadata_request.headers
+            await metadata_request.release()
+            if request_code != 200:
+                raise exceptions.MetadataError('Failed to fetch file metadata from WaterButler. Received response: code {} {}'.format(str(request_code), str(request_reason)),
```
cslzchen: Similar as the comment above (actual code is below).

cslzchen: Do we want this exception caught by sentry? Will it be caught automatically? (I came from OSF where we need to explicitly call sentry to log exceptions ...)

mfr/providers/osf/provider.py
```diff
@@ -66,23 +66,24 @@ def __init__(self, request, url):
...
                     metadata_url=download_url,
-                    response=await metadata_request.text(),
+                    response=request_reason,
                     provider=self.NAME,
-                    code=400,
-                )
+                    code=400)
```
cslzchen: One for `200` and one for `400`. It will be great if we can gradually replace number literals with status constants in our code base ([link](https://docs.python.org/3.5/library/http.html#http.HTTPStatus)). For example:
```python
from http import HTTPStatus

httpStatus.OK
httpStatus.BAD_REQUEST
```

mfr/providers/osf/provider.py
```diff
@@ -66,23 +66,24 @@ def __init__(self, request, url):
             metadata_url = download_url.replace('/file?', '/data?', 1)
             metadata_request = await self._make_request('GET', metadata_url)
```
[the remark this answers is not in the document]
TomBaxter: I see what you mean. I could see it either way. It returns a request object, but it also represents the response to an HTTP request.

mfr/providers/osf/provider.py
```diff
@@ -66,23 +66,24 @@ def __init__(self, request, url):
...
             metadata = await metadata_request.json()
+            await metadata_request.release()
```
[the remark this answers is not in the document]
TomBaxter: It's been a while. I think I'm going to be saying that to you a lot now that you've taken on these ancient PRs. Anyway, I believe this is simply because it has become our standard to manually close all requests. I believe this is because we were getting a ton of "unclosed" type errors in sentry. @felliott correct me if I'm wrong here.

mfr/providers/osf/provider.py
TomBaxter: Automatically, I believe
https://github.com/CenterForOpenScience/modular-file-renderer/blob/develop/mfr/server/handlers/core.py#L154

mfr/providers/osf/provider.py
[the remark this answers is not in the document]
cslzchen: Np, I understand. That's why I try to ask as many questions to help us recall the decision made long time ago, double check a change and document it somewhere. Hope it is constructive rather than causing overhead for you guys.

I agree that it is better we explicitly close/release the connection.

mfr/providers/osf/provider.py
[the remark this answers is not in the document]
cslzchen: `._make_request()` calls `aiohttp.request()` which returns a response object: [link](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L739). I am ok calling it request. If we were to change it, it would be everywhere I think.

mfr/providers/osf/provider.py
TomBaxter: It's expecting text. I'm afraid I can't recall why I changed it, and I don't see where the exception actually does anything with it.

mfr/providers/osf/provider.py
TomBaxter: Seems like a good idea.

mfr/providers/osf/provider.py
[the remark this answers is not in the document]
TomBaxter: Change made.

mfr/providers/osf/provider.py
```diff
@@ -64,25 +64,26 @@ def __init__(self, request, url):
...
-            except KeyError:
-                raise exceptions.MetadataError(
-                    'Failed to fetch metadata. Received response '
-                    'code {}'.format(str(metadata_request.status)),
+            metadata_response = await self._make_request('HEAD', download_url)
+            response_code = metadata_response.status
+            response_reason = metadata_response.reason
+            response_headers = metadata_response.headers
+            await metadata_response.release()
+            if response_code != 200:
```
cslzchen: Line 79 is too long, how about his? (Not a blocker)
```python
if response_code != 200:
    raise exceptions.MetadataError(
        'Failed to fetch file metadata from WaterButler. Received response: code {} {}'.format(
            str(response_code), str(response_reason)
        ),
        metadata_url=download_url,
        response=response_reason,
        provider=self.NAME,
        code=400
    )
```

mfr/providers/osf/provider.py
```diff
@@ -64,25 +64,26 @@ def __init__(self, request, url):
...
+                raise exceptions.MetadataError('Failed to fetch file metadata from WaterButler. Received response: code {} {}'.format(str(response_code), str(response_reason)),
                     metadata_url=download_url,
-                    response=await metadata_request.text(),
+                    response=response_reason,
```
cslzchen: In case you had wanted to change to `.text()` but forgot. (I am good with either `.text()` or `.reason`. (Not a blocker)

mfr/providers/osf/provider.py
cslzchen: There are too many status in the file. No need to change for now. Maybe in the future we can have a clean ticket for this but very low priority for now.

felliott:
Moving to RTM, will be in v0.23

PR #266, opened 2017-06-21
[SVCS-287] Make Iframes scrollable for IE11

Johnetordoff:
## Purpose

The MFR Iframe currently isn't scrollable in IE11 this fix solves that,

## Change

Changes iframe attribute to allow scrolling.

## Side Effect

None that I know of

## Ticket

https://openscience.atlassian.net/browse/SVCS-287

### DIFF
--- a/mfr/server/static/js/mfr.js
+++ b/mfr/server/static/js/mfr.js
@@ -80,6 +80,8 @@
             self.pymParent.iframe.setAttribute('allowfullscreen', '');
             self.pymParent.iframe.setAttribute('webkitallowfullscreen', '');
             self.pymParent.iframe.setAttribute('scrolling', 'yes');
+            self.pymParent.iframe.setAttribute('horizontalscrolling', 'yes');
+            self.pymParent.iframe.setAttribute('verticalscrolling', 'yes');
 
             self.pymParent.el.appendChild(self.spinner);
             $(self.pymParent.iframe).load(function () {


mfr/server/static/js/mfr.js
```diff
@@ -80,6 +80,8 @@
             self.pymParent.iframe.setAttribute('allowfullscreen', '');
             self.pymParent.iframe.setAttribute('webkitallowfullscreen', '');
             self.pymParent.iframe.setAttribute('scrolling', 'yes');
+            self.pymParent.iframe.setAttribute('horizontalscrolling', 'yes');
```
cslzchen: 1. The `scrolling` attribute for `iframe` is not supported by HTML5, can this be the cause? It seems IE11 dropped several support for `html4` to better support `html5` and `css`: https://msdn.microsoft.com/en-us/library/bg182625(v=vs.85).aspx.
2. Is `horizontalscrolling` and `verticalscrolling` a valid `iframe` attributes? I can't find it though.

cslzchen: 1. The `scrolling` attribute for `iframe` is not supported by `HTML5`, can this be the reason? It seems that IE11 dropped several support for `HTML4` to better support `HTML5` and `CSS`: https://msdn.microsoft.com/en-us/library/bg182625(v=vs.85).aspx.
2. Is `horizontalscrolling` and `verticalscrolling` a valid `iframe` attributes? I can't find it though.

Johnetordoff: I don't know about the documentation or how exactly this is supported, but it does work. I tested locally in Virtual Box test environment provided by Microsoft. I would also create another ticket for the Pym update.

cslzchen: Sounds good and thanks. The commit message is good enough for us to keep track of why we made this change. 👍

cslzchen:
Looks good. Move to PCR. 🎆 🔥 🚒 👩‍🚒 👨‍🚒 
(I didn't test this locally since I don't have the windows virtual machine setup.)

PR #268, opened 2017-07-26
[SVCS-404] Added support for rendering mat files

AddisonSchiller:
https://openscience.atlassian.net/browse/SVCS-404

# Purpose
add the .mat format to tabular file renderer 

# Summary of Changes
Added a new file under tabular/libs which contains functions to render the .mat format.
The .mat format contains variables from matlab scripts. Each variable is saved as a sheet (wether it be 1 number, or a 100x100 matrix) and then can be viewed.  Two libraries were added for this. h5py and scipy.

# Testing Notes

There are 8 files on the jira ticket that can be used for testing. Files ending in V73 test the V7.3 h5py renderer. Files ending in V70 test the V7 and below scipy renderer. The files come in pairs, one for the V73 and one for the V70. Files across both versions should have the same contents. However the sheets may not be listed in the same order across both versions.


mfr/extensions/tabular/libs/h5py_scipy_tools.py
```diff
@@ -0,0 +1,60 @@
...
+        list_data = workbook[key].tolist()
+        # add default header to fix rendering of some data formats
+        header = [i for i in range(1, 1 + len(list_data[0]))]
+        values = data_population(list_data, headers=header)
+        header = header_population(header)
+        sheets[str(key)] = (header, values)
+
+    return sheets
+
+def mat_h5py_scipy(fp):
```
cslzchen: 1. If an exception is expected, we can to document it using `:raises`.
2. For `try ... except ...`, what is the exact exception(s) that a `v7` parser throws on a `v73` file? We should be specific on exceptions.
3. **OPTIONAL** Is there alternative ways to identify the version other than exceptions?

cslzchen: How about malformed/corrupted files? I guess that some exceptions are thrown. We should catch and handle them in `mat_v7()` or `mat_v73()` instead of throwing `500`. The front end can inform the user that the file cannot be rendered due to corrupted file or unsupported version.

Please note that this may applies for all types of files. We may need another clean-up ticket if we don't already handle corrupted files as a standard, which is out of the scope of this ticket. However, it is better to have this check for new ones we are adding. Not a blocker though.

mfr/extensions/tabular/libs/h5py_scipy_tools.py
```diff
@@ -0,0 +1,60 @@
...
+from collections import OrderedDict
+
+
+def mat_v73(fp):
+    """Read and convert a mat v7.3+ file to JSON format using h5py
+    :param fp: File pointer object
+    :return: tuple of table headers and data
+    """
+
+    workbook = h5py.File(fp.name, 'r')
```
cslzchen: As mentioned below, please check if this throws a certain exception and if `workbook` is `None`. Here is an example structure I have in mind just to provide some insights.

```python
try:
     workbook = h5py.File(fp.name, 'r')
except ...:
    # invalid version, re-raise it for trying another parser
     ...
except ...:
    # other parsing failure
    # log the issue and set workbook to None
    ...

if workbook:   
    # your code
    sheets = OrderedDict()
    variables = workbook.items()
    ...
    return sheets

# return an error message back to the user (not sure if it is here or handled by something else)
```

mfr/extensions/tabular/libs/h5py_scipy_tools.py
```diff
@@ -0,0 +1,60 @@
...
+    sheets = OrderedDict()
+    variables = workbook.items()
+
+    for var in variables:
+        name = var[0]
+        data = var[1]
+        if type(data) is h5py.Dataset:
+            #h5py seems to swap cols/rows for some reason. this fixes it so it displays like it does in matlab
+            list_data = list(zip(*data.value.tolist()))
+            #add a default header to fix some data formats
```
cslzchen: Can we make this a helper method since it is used by both `v73` and `v7`?

mfr/extensions/tabular/libs/h5py_scipy_tools.py
```diff
@@ -0,0 +1,60 @@
...
+
+    workbook = h5py.File(fp.name, 'r')
+    sheets = OrderedDict()
+    variables = workbook.items()
+
+    for var in variables:
+        name = var[0]
+        data = var[1]
+        if type(data) is h5py.Dataset:
+            #h5py seems to swap cols/rows for some reason. this fixes it so it displays like it does in matlab
```
cslzchen: Yes. HDF5 always stores data as row-major (C convention).

mfr/extensions/tabular/libs/h5py_scipy_tools.py
cslzchen: And it will be great if you can add a comment/explanation on the `list_data` format. Not a blocker.

tests/extensions/tabular/test_mat_tools.py
```diff
@@ -0,0 +1,25 @@
...
+from mfr.extensions.tabular.libs import mat_h5py_scipy
+
+BASE = os.path.dirname(os.path.abspath(__file__))
+
+
+class TestTabularMatTools:
+
+    def test_mat_73(self):
+        with open(os.path.join(BASE, 'files', 'testVer73.mat')) as fp:
+            # difference is with actual file, not function
```
cslzchen: I don't think this comment is necessary since the function name `mat_h5py_scipy()` is good enough. (2 occurrences)

mfr/extensions/tabular/libs/h5py_scipy_tools.py
```diff
@@ -0,0 +1,75 @@
...
+    try:
+        workbook = scipy.io.loadmat(fp.name)
+    except Exception as e:
+        if type(e) is NotImplementedError:
+            raise e
+        else:
+            raise(exceptions.EmptyTableError("Cannot render an empty mat file"))
+
+    if not workbook:
+        raise(exceptions.EmptyTableError("Cannot render an empty mat file"))
```
cslzchen: Just checked the `scipy.io.loadmat()` [doc](https://docs.scipy.org/doc/scipy-0.19.1/reference/generated/scipy.io.loadmat.html), it always returns an dictionary, can be `empty` but never `None`. So instead of "Cannot render an empty mat file", can we use something like "Warning: the .mat file is empty"?

mfr/extensions/tabular/libs/h5py_scipy_tools.py
```diff
@@ -0,0 +1,75 @@
...
+def mat_v73(fp):
+    """Read and convert a mat v7.3+ file to JSON format using h5py
+    :param fp: File pointer object
+    :return: tuple of table headers and data
+    """
+
+    workbook = h5py.File(fp.name, 'r')
+
+    if not workbook:
+        raise(exceptions.EmptyTableError("Cannot render an empty mat file"))
```
cslzchen: Similarly here, `h5py.File()` [doc](http://docs.h5py.org/en/latest/high/file.html#File) is a constructor so it will either raise an exception or return an object which can never be `None`. We can use something like "Warning: the .mat file is empty" as mentioned below.

mfr/extensions/tabular/libs/h5py_scipy_tools.py
```diff
@@ -0,0 +1,75 @@
...
+    :param fp: File pointer object
+    :return: tuple of table headers and data
+    """
+    try:
+        workbook = scipy.io.loadmat(fp.name)
+    except Exception as e:
+        if type(e) is NotImplementedError:
+            raise e
+        else:
+            raise(exceptions.EmptyTableError("Cannot render an empty mat file"))
```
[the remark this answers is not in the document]
AddisonSchiller: Ah yes, i added it when writing tests and just followed the other examples we talked about. It definitely makes more sense to use a different error however.

tests/extensions/tabular/test_mat_tools.py
AddisonSchiller: I think if someone was looking at this and didn't know how it worked, the comment could be helpful. Maybe I should elaborate on it a bit more?

cslzchen: Sure, let's elaborate it. (I was confused the first time when I read it, until I see the function name lol.)

mfr/extensions/tabular/libs/h5py_scipy_tools.py
```diff
@@ -0,0 +1,72 @@
+from collections import OrderedDict
+
+import h5py
+import scipy.io
+
+from mfr.extensions.tabular import exceptions
+from mfr.extensions.tabular.utilities import data_population, header_population
+
```
cslzchen: Two empty lines.

mfr/extensions/tabular/libs/h5py_scipy_tools.py
```diff
@@ -0,0 +1,72 @@
...
+def build_sheets(name, list_data, sheets):
+    # add default header to fix rendering of some data formats
+    header = [i for i in range(1, 1 + len(list_data[0]))]
+    values = data_population(list_data, headers=header)
+    header = header_population(header)
+    sheets[str(name)] = (header, values)
+    return sheets
+
```
cslzchen: Two empty lines.

mfr/extensions/tabular/libs/h5py_scipy_tools.py
```diff
@@ -0,0 +1,72 @@
...
+
+def mat_v73(fp):
+    """Read and convert a mat v7.3+ file to JSON format using h5py
+    :param fp: File pointer object
+    :return: tuple of table headers and data
+    """
+
+    # workbook cannot be empty or none. Will atleast have some sort of data in it
```
cslzchen: "at least"

mfr/extensions/tabular/libs/h5py_scipy_tools.py
```diff
@@ -0,0 +1,72 @@
...
+    :param fp: File pointer object
+    :return: tuple of table headers and data
+    """
+    try:
+        # workbook cannot be empty or none, will always have header information
+        workbook = scipy.io.loadmat(fp.name)
+    except NotImplementedError as e:
+        raise e
+    except Exception:
+        raise(tabular_exceptions.UnexpectedFormattingError(
```
cslzchen: No need to use `()` for `raise`.

tests/extensions/tabular/test_renderer.py
```diff
@@ -46,6 +46,25 @@ def invalid_xlsx_file_path():
 
 
 @pytest.fixture
+def invalid_mat73_file_path():
+    return os.path.join(os.path.dirname(os.path.abspath(__file__)), 'files', 'invalidVer73.mat')
```
felliott: While you're in here, how about saving `os.path.dirname(os.path.abspath(__file__)), 'files'` into an intermediate variable, so save all the c-and-p in the fixtures?

tests/extensions/tabular/test_renderer.py
```diff
@@ -117,3 +136,30 @@ def test_render_tabular_xlsx_invalid(self, invalid_xlsx_file_path, assets_url, e
         renderer = TabularRenderer(metadata, invalid_xlsx_file_path, url, assets_url, export_url)
         with pytest.raises(XLRDError):
             renderer.render()
+
+
+class TestTabularMatRenderer:
+
+    def test_render_tabular_mat70(self, test_mat70_file_path, assets_url, export_url):
+        metadata = ProviderMetadata('test', '.mat', 'text/plain', '1234', 'http://wb.osf.io/file/testVer70.mat?token=1234')
```
felliott: Line length

tests/extensions/tabular/test_renderer.py
```diff
@@ -117,3 +136,30 @@ def test_render_tabular_xlsx_invalid(self, invalid_xlsx_file_path, assets_url, e
...
+        renderer = TabularRenderer(metadata, test_mat70_file_path, url, assets_url, export_url)
+        body = renderer.render()
+        assert '<div id="mfrGrid" style="min-height: {}px;">\n    </div>'.format(settings.TABLE_HEIGHT) in body
```
felliott: line length, and if this is a common testing var (it's used again in `test_render_tabular_mat73`) maybe make it a fixture or global.

mfr/extensions/tabular/libs/h5py_scipy_tools.py
```diff
@@ -0,0 +1,72 @@
...
+        workbook.pop(key, None)
+
+    for key in workbook:
+        list_data = workbook[key].tolist()
+        sheets = build_sheets(key, list_data, sheets)
+
+    return sheets
+
+def mat_h5py_scipy(fp):
+    try:
```
felliott: Please add a blurb explaining what you're doing (trying to load a file w/ two different libraries), why you're doing it (h5py doesn't support <v7.3, scipy doesn't support >v7.0), and why you try one before the other (even if it's arbitrary).

In a related note, am I misunderstanding the naming or does this not support mat files v7.1 and v7.2?  If so, that's weird enough that it might be worth a bespoke error message.

mfr/extensions/tabular/libs/h5py_scipy_tools.py
```diff
@@ -0,0 +1,72 @@
...
+        raise e
+    except Exception:
+        raise(tabular_exceptions.UnexpectedFormattingError(
+            'Cannot render this file at this time.'
+            ' The file may not be a .mat file, or may be corrupted'))
+
+    sheets = OrderedDict()
+
+    for key in ('__header__', '__version__', '__globals__'):
+        workbook.pop(key, None)
```
felliott: Totally minor, but why bother mutating the workbook, instead of just skipping the offending keys?

tests/extensions/tabular/test_mat_tools.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def test_mat_70(self):
+        with open(os.path.join(BASE, 'files', 'testVer70.mat')) as fp:
+            # The two test functions are basically exactly the same. That is because
+            # there are different types of mat files. The only way to tell them apart is to
+            # try and load them.
+            sheets = mat_h5py_scipy()(fp)
+
+        sheet = sheets.popitem()[1]
+        print(sheet)
```
felliott: Sinner.

mfr/extensions/tabular/libs/h5py_scipy_tools.py
AddisonSchiller: http://www.mathworks.com/help/matlab/import_export/mat-file-versions.html 
From my understanding there is no 7.1 or 7.2 (there may be a 7.1, but from what I've read scipy can open it if it exists).

The file types use different compressions and can store differing amounts of data. It seems they are just named after whatever matlab version they are released with (? just a guess).

tests/extensions/tabular/test_mat_tools.py
AddisonSchiller: I feel a little bad, but at least it was in a test

mfr/extensions/tabular/libs/h5py_scipy_tools.py
AddisonSchiller: Was just the way my thought process went about it at the time. Will change.

felliott:
Ace!  Moving to RTM!

PR #271, opened 2017-08-03
[SVCS-403]fixed support for .dta file extension

AddisonSchiller:
https://openscience.atlassian.net/browse/SVCS-403

# Purpose
Investigate previously added MFR extensions. Fix outdated ones and re add them. remove ones that do not work properly.
# Summary of Changes
Added support back in for .dta file type. removed support for swf (was rendering via unoconv, when it is an adobe flash file.)  removed mentions of .avi since it is not supported

# Testing Notes

There are a few files on the jira ticket for testing. All should render properly except the error.dta file. That file is for testing to make sure the correct error is thrown when trying a render a file  non dta file as dta. Columns should not have values show up as NaN, instead they should show up as blank.


mfr/extensions/tabular/render.py
```diff
@@ -80,6 +80,12 @@ def _populate_data(self, fp, ext):
                         extension=self.metadata.ext,
                         formatting_function=str(function),
                     )
+                except ValueError:
```
cslzchen: There are two types of exceptions here: `KeyError` and `ValueError`. I am able to trigger the latter by using the `error.dta`.

1. Is `ValueError` only applicable to `.dta` files? We should have a more general message saying the file can't be rendered. Instead of hard-coding `as .dta`, we can use a format string.
2. If both `KeyError` and `ValueError` are triggered by invalid file format, we can combine  them together.

cslzchen:
More on exception handling:

```python
LIBS = config.get('LIBS', {
    '.csv': [libs.csv_stdlib],
    '.tsv': [libs.csv_stdlib],
    '.gsheet': [libs.xlsx_xlrd],
    '.xlsx': [libs.xlsx_xlrd],
    '.xls': [libs.xlsx_xlrd],
    '.dta': [libs.dta_pandas],
    '.sav': [libs.sav_stdlib],
    # '.ods': [libs.ods_ezodf],
})
```

All these types uses the same render and shares exception handling. Can you test them all with malformed files and see what exceptions are thrown?

setup.py
```diff
@@ -736,7 +734,6 @@ def parse_requirements(requirements):
             # video
             '.mp4 = mfr.extensions.video:VideoRenderer',
             '.m4v = mfr.extensions.video:VideoRenderer',
-            #'.avi = mfr.extensions.video:VideoRenderer',
```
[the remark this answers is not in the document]
AddisonSchiller: These files were ones that were commented out, or currently rendered (but very buggy and or producing junk). I determined that .avi files could not currently be rendered, so I just removed the comment. Maybe it is best to just leave it commented out for historic reasons?

mfr/extensions/tabular/render.py
AddisonSchiller: Going with #2 here, seems to make the most sense to me here. `error.dta` was a renamed csv. So it really is a formatting error.

tests/extensions/tabular/files/invalid.dta
```diff
@@ -0,0 +1,8 @@
+,id,target,gender,likely_discriminated,promoted,responsible,compensated,would_apply,chances,attractive_employer,put_off_similar_jobs,work,work_experience,studies,age
```
cslzchen: Is this file used anywhere in tests?

tests/extensions/tabular/test_panda_tools.py
```diff
@@ -31,11 +31,30 @@ def test_tsv_pandas(self):
...
 
-    # def test_dta_pandas():
-    #     with open('mfr_tabular/tests/fixtures/test.dta') as fp:
-    #         headers, data = panda_tools.dta_pandas(fp)
-
-    #     assert headers[0] == {'field': 'one', 'id': 'one', 'name': 'one'}
-    #     assert data[0] == {'one': 'a', 'two': 'b', 'three': 'c'}
-    #     assert len(data) is 2
-    #     assert len(headers) is 3
+    def test_dta_pandas(self):
```
cslzchen: Let's add a test for invalid `.dta` files.

PR #273, opened 2017-08-18
[SVCS-423] Links in pdfs should no longer open in the iframe.

AddisonSchiller:
Instead, they should open in a new tab/window

https://openscience.atlassian.net/browse/SVCS-423

## Purpose
Fix opening links in an iframe.
Update tests to match fix.

## Summary of changes
added default behavior of links to open in new tab, instead of on same page (in iframe)
Changed MFR tests to match this behavior.

## QA Notes
Open a pdf that has a link in it (example on jira ticket)
Click on the link
It should redirect you in a new tab, instead of in the iframe.

Mfr pdf tests should pass

### DIFF
--- a/mfr/extensions/pdf/templates/viewer.mako
+++ b/mfr/extensions/pdf/templates/viewer.mako
@@ -19,7 +19,7 @@ http://sourceforge.net/adobe/cmap/wiki/License/
 -->
 <html dir="ltr" mozdisallowselectionprint moznomarginboxes>
   <head>
-    <base href="${base}/web/">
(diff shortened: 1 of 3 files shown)


AddisonSchiller:
fixed. Sorta killed the commit for it when merging. Anyway, sandboxing just needed 'allow-popups'.

https://stackoverflow.com/questions/21236616/sandboxed-iframe-with-target-blank-doesnt-open-new-tab-or-window

felliott:
yaaaaaaaaaaaaaaaaaaaay. merged.

PR #275, opened 2017-08-31
[SVCS-52] Adding support for mutli-page tiff files.

AddisonSchiller:
adding more to pdf renderer
tiff files are now rendered by the pdf renderer, and the pdf extension now has an exporter to handle tif/tiff files. Tiff files are now converted to pdfs for viewing

refs: https://openscience.atlassian.net/browse/SVCS-52

## Purpose
Add support for mutlipage tiff files.

## Summary of changes
Tiff files are sort of like pdfs, and can be converted to them. 
The pdf extension now handles tiff files. The pdf extension now has an exporter for tiff files
Reportlab has been added as a dependency. 
Some multipage tiffs will not render correctly. see testing notes. 

## Testing QA notes:
below is a copy and paste of my testing notes on the jira ticket. you can find the zip file over there

Testing Notes:

Attached to the jira ticket is a zip containing some tif/tiff files to use for testing.

Due to the aspects of what this commit is doing, the following should also be tested:
    Other pdf files. Try a few regular pdf files and make sure they render properly.
    Any files rendered by the unoconv renderer. (ie docx) The unoconv renderer ships files to the pdf renderer.


mfr/extensions/pdf/export.py
```diff
@@ -0,0 +1,85 @@
...
+        page = 0
+
+        # This seems to be the only way to write this loop at the moment
+        while True:
+            try:
+                tiff_img.seek(page)
+            except EOFError:
+                break
+
+            # Center the image and draw it in the canvas
```
cslzchen: Be consistent on comments style (a few occurrences).
* should we capitalize the first letter of the first word (whichever you prefer)
* should we add punctuation at the end of the line (usually not)

mfr/extensions/pdf/export.py
```diff
@@ -0,0 +1,85 @@
...
+            if max_size:
+                # resize the image to the w/h maximum specified
+                ratio = min(max_size[0] / image.size[0], max_size[1] / image.size[1])
+                self.metrics.add('ratio', ratio)
+                if ratio < 1:
+                    image = image.resize((round(image.size[0] * ratio), round(image.size[1] * ratio)), Image.ANTIALIAS)
+
+            self.tiff_to_pdf(image, max_size)
+            image.close()
+
```
cslzchen: Remove this empty line.

mfr/extensions/pdf/export.py
```diff
@@ -0,0 +1,85 @@
...
+            name, extension = os.path.splitext(os.path.split(self.source_file_path)[-1])
+            raise exceptions.PillowImageError(
+                'Unable to export the file as a {}, please check that the '
+                'file is a valid image.'.format(type),
+                export_format=type,
+                detected_format=imghdr.what(self.source_file_path),
+                original_exception=err,
+                code=400,
+            )
+        except(ValueError) as err:
```
cslzchen:
```python
except ValueError as err:
```

mfr/extensions/pdf/export.py
```diff
@@ -0,0 +1,85 @@
...
+                # resize the image to the w/h maximum specified
+                ratio = min(max_size[0] / image.size[0], max_size[1] / image.size[1])
+                self.metrics.add('ratio', ratio)
+                if ratio < 1:
+                    image = image.resize((round(image.size[0] * ratio), round(image.size[1] * ratio)), Image.ANTIALIAS)
+
+            self.tiff_to_pdf(image, max_size)
+            image.close()
+
+        except (UnicodeDecodeError, IOError) as err:
```
cslzchen: The code and exception message looks similar. Why not combining the two `except` in one? (A question not a suggestion.)

mfr/extensions/pdf/export.py
```diff
@@ -0,0 +1,85 @@
...
+            name, extension = os.path.splitext(os.path.split(self.source_file_path)[-1])
+            raise exceptions.PillowImageError(
+                'Unable to export the file as a {}, please check that the '
+                'file is a valid image.'.format(type),
+                export_format=type,
+                detected_format=imghdr.what(self.source_file_path),
+                original_exception=err,
+                code=400,
```
cslzchen: Given this is a new file, let's use `HTTPStatus` ([LINK](https://docs.python.org/3/library/http.html#http-status-codes)).
```python
from http import HTTPStatus
```
```python
code=HTTPStatus.BAD_REQUEST
```

mfr/extensions/pdf/render.py
```diff
@@ -13,7 +15,19 @@ class PdfRenderer(extension.BaseRenderer):
...
+        if self.metadata.ext in settings.EXPORT_EXCLUSIONS:
+            return self.TEMPLATE.render(base=self.assets_url, url=self.metadata.download_url)
+
+        exported_url = furl.furl(self.export_url)
+        if settings.EXPORT_MAXIMUM_SIZE and settings.EXPORT_TYPE:
+            exported_url.args['format'] = '{}.{}'.format(settings.EXPORT_MAXIMUM_SIZE, settings.EXPORT_TYPE)
+        elif settings.EXPORT_TYPE:
+            exported_url.args['format'] = settings.EXPORT_TYPE
+        else:
+            return self.TEMPLATE.render(base=self.assets_url, url=self.metadata.download_url)
```
cslzchen: What's the conditions for reaching this branch? Is this unexpected?

mfr/extensions/pdf/render.py
```diff
@@ -13,7 +15,19 @@ class PdfRenderer(extension.BaseRenderer):
         ]).get_template('viewer.mako')
 
     def render(self):
-        return self.TEMPLATE.render(base=self.assets_url, url=self.metadata.download_url)
+        if self.metadata.ext in settings.EXPORT_EXCLUSIONS:
+            return self.TEMPLATE.render(base=self.assets_url, url=self.metadata.download_url)
+
+        exported_url = furl.furl(self.export_url)
+        if settings.EXPORT_MAXIMUM_SIZE and settings.EXPORT_TYPE:
```
cslzchen: This reduce the # of `bool` checks from 3 to 2.
```python
if settings.EXPORT_TYPE:
    if settings.EXPORT_MAXIMUM_SIZE
        exported_url.args['format'] = '{}.{}'.format(settings.EXPORT_MAXIMUM_SIZE, settings.EXPORT_TYPE)
    else:
        exported_url.args['format'] = settings.EXPORT_TYPE
    self.metrics.add('needs_export', True)
    return self.TEMPLATE.render(base=self.assets_url, url=exported_url.url)

# TODO: is this an unexpected case?
return self.TEMPLATE.render(base=self.assets_url, url=self.metadata.download_url)
```

mfr/extensions/pdf/exceptions.py
```diff
@@ -0,0 +1,23 @@
+from mfr.core.exceptions import ExporterError
+
+
+class PillowImageError(ExporterError):
```
cslzchen: Seems duplicated, but I guess you have a good reason for not using `mfr.extensions.image.exceptions.PillowImageError`?

AddisonSchiller: It is duplicated, but it doesnt make sense to import an exception from a different extension (for the most part, extensions dont interact much) thats why its copied this way. Thoughts?

mfr/extensions/pdf/render.py
AddisonSchiller: This was mostly copied from the image renderer. I assume that the TODO case here is when the settings are not properly defined. Going with your approach here but leaving the unexpected case in for completion etc.

mfr/extensions/pdf/export.py
AddisonSchiller: as a note, the `ValueError` case is a bug with our current version of pillow having trouble opening certain formats of `.tiff`. Once pillow is updated to 4.3 or higher, this last error section can go away.

mfr/extensions/pdf/export.py
AddisonSchiller: The `ValueError` case is a bug with our current version of pillow having trouble opening certain formats of `.tiff`. Once pillow is updated to 4.3 or higher, this last error section can go away. (tested now and this is indeed the case. `broken.tiff` will open on pillow 4.3, but not 2.8.3 that we currently have. This is why the errors were separate. The tiff given was valid, just pillow messing up.

mfr/extensions/pdf/export.py
```diff
@@ -0,0 +1,85 @@
...
+
+            if max_pages and page > max_pages:
+                break
+            page += 1
+
+        c.save()
+
+    def export(self):
+        parts = self.format.split('.')
+        type = parts[-1].lower()
```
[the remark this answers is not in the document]
AddisonSchiller: This code is directly from the image  exporter.

mfr/extensions/pdf/export.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def __init__(self, *args, **kwargs):
+
+        super().__init__(*args, **kwargs)
+        self.metrics.add('pil_version', Image.VERSION)
+
+    def tiff_to_pdf(self, tiff_img, max_size):
+        max_pages = 40
+        width, height = tiff_img.size
+        c = canvas.Canvas(self.output_file_path)
```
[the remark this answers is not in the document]
AddisonSchiller: There should be no file at output_file_path. Thats the path MFR wants the exported file to be saved to.

mfr/extensions/pdf/__init__.py
```diff
@@ -1 +1,2 @@
 from .render import PdfRenderer  # noqa
+from .export import PdfExporter # noqa
```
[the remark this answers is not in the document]
AddisonSchiller: Leaving this like we talked about on other issue

mfr/extensions/pdf/export.py
```diff
@@ -0,0 +1,85 @@
+import os
+
+import imghdr
+from PIL import Image
+
+from reportlab.pdfgen import canvas
+
+from mfr.core import extension
+from mfr.extensions.pdf import exceptions
```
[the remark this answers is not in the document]
AddisonSchiller: Imghdr is a base lib, so leaving it with os, moving others.

mfr/extensions/pdf/export.py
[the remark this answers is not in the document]
cslzchen: Ah, I see. I thought there is a temp file created.

mfr/extensions/pdf/export.py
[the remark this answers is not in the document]
cslzchen: Let's change this here. Using reserved word is a huge issue. If possible (not causing too much trouble), update the image exporter as well. Thanks.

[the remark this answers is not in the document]
AddisonSchiller: I will update the image exporter on the ticket I just did for it. 👍

mfr/extensions/pdf/render.py
cslzchen: Agreed. Similarly, update image renderer if possible.

mfr/extensions/pdf/exceptions.py
cslzchen: I see what you mean and I agree. Let's leave this here for @felliott in Phase 2 Review.

felliott: I'm okay with duplicating this, but the docstrings and `__TYPE` should be consistent with the new location.  `__TYPE` should probably be `pdf_pillow`.

mfr/extensions/pdf/render.py
```diff
@@ -13,6 +15,20 @@ class PdfRenderer(extension.BaseRenderer):
...
 
     def render(self):
+        if self.metadata.ext in settings.EXPORT_EXCLUSIONS:
+            return self.TEMPLATE.render(base=self.assets_url, url=self.metadata.download_url)
+
+        exported_url = furl.furl(self.export_url)
+        if settings.EXPORT_TYPE:
+            if settings.EXPORT_MAXIMUM_SIZE:
+                exported_url.args['format'] = '{}.{}'.format(settings.EXPORT_MAXIMUM_SIZE,
+                                                                    settings.EXPORT_TYPE)
```
felliott: nitpick: indentation

mfr/extensions/pdf/settings.py
```diff
@@ -0,0 +1,9 @@
+from mfr import settings
+
+
+config = settings.child('PDF_EXTENSION_CONFIG')
+
+EXPORT_TYPE = config.get('EXPORT_TYPE', 'pdf')
+EXPORT_MAXIMUM_SIZE = config.get('EXPORT_MAXIMUM_SIZE', '1200x1200')
+EXPORT_EXCLUSIONS = config.get('EXPORT_EXCLUSIONS', ['.pdf', ])
+EXPORT_MAX_PAGES = config.get('EXPORT_MAX_PAGES', 40)
```
felliott: The above two config settings will have problems if set from the env.  Envvars are always strings; they won't be ints or lists.  I believe there are some helper methods in the config base class to cast these properly.

tests/extensions/pdf/test_exporter.py
```diff
@@ -0,0 +1,93 @@
+import os
+import shutil
+
+import pytest
+
+from mfr.extensions.pdf import settings
+from mfr.extensions.pdf import exceptions
+from mfr.extensions.pdf import PdfExporter
```
felliott: The above could be collapsed.

tests/extensions/pdf/test_exporter.py
```diff
@@ -0,0 +1,93 @@
...
+
+@pytest.fixture(scope="function", autouse=True)
+def setup_filesystem(directory):
+    shutil.rmtree(directory)
+    os.makedirs(directory, exist_ok=True)
+
+
+class TestPdfExporter:
+    '''
+    Opening and veryifying pdfs with report lab is a paid feature
```
felliott: Nitpick: "verifying".  Also, "report lab" could probably use a little more context here, it's only mentioned/used in requirements.txt and the renderer file.

tests/extensions/pdf/test_exporter.py
```diff
@@ -0,0 +1,93 @@
...
+    so these tests are mostly just to see if the file gets created
+    in the correct place
```
felliott: Is there anything else in the stdlib that can assert that the output is actually a pdf?  If not, no worries, it would just be good to assert that a pdf is produced, rather than e.g. and html error message.

felliott: Followup: it looks like PDFs have a signature: https://en.wikipedia.org/wiki/List_of_file_signatures

tests/extensions/pdf/test_exporter.py
```diff
@@ -0,0 +1,93 @@
...
+        output_file_path = os.path.join(directory, 'test.{}'.format(settings.EXPORT_TYPE))
+        format = '{}.{}'.format(settings.EXPORT_MAXIMUM_SIZE, settings.EXPORT_TYPE)
+        exporter = PdfExporter(source_file_path=source_file_path,
+                            output_file_path=output_file_path, format=format)
+
+        assert not os.path.exists(output_file_path)
+        assert os.path.exists(directory)
+
+        exporter.export()
+        assert os.path.exists(output_file_path)
```
felliott: It looks like the successful tests and the failing tests could be parameterized, since we don't have the ability to do deep examination of the pdfs.

mfr/extensions/pdf/settings.py
```diff
@@ -0,0 +1,9 @@
+from mfr import settings
+
+
+config = settings.child('PDF_EXTENSION_CONFIG')
+
+EXPORT_TYPE = config.get('EXPORT_TYPE', 'pdf')
+EXPORT_MAXIMUM_SIZE = config.get('EXPORT_MAXIMUM_SIZE', '1200x1200')
+EXPORT_EXCLUSIONS = config.get('EXPORT_EXCLUSIONS', ['.pdf', ])
```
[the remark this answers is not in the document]
AddisonSchiller: the image settings is done the same way (its mostly c/p from the image settings). I don't see a method in mfr.settings to handle lists. 
Should I be feeding this a string and making the list myself in the renderer or something like that?

also, should the image extension be changed as well?

mfr/extensions/pdf/settings.py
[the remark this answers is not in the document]
AddisonSchiller: complete

mfr/extensions/pdf/exceptions.py
```diff
@@ -0,0 +1,23 @@
...
+    """The Image related errors raised from a :class:`mfr.extentions.image`
+    and relating to the Pillow Library should inherit from PillowImageError
+    """
+
+    __TYPE = 'image_pillow'
```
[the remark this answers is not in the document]
AddisonSchiller: complete

tests/extensions/pdf/test_exporter.py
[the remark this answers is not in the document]
AddisonSchiller: complete

tests/extensions/pdf/test_exporter.py
AddisonSchiller: This is great, implementing

tests/extensions/pdf/test_exporter.py
AddisonSchiller: What do you mean by this?

tests/extensions/pdf/test_renderer.py
```diff
@@ -1,26 +1,41 @@
...
+from mfr.extensions.pdf import settings
 from mfr.extensions.pdf import PdfRenderer
+from mfr.core.provider import ProviderMetadata
 
 
 @pytest.fixture
 def metadata():
-    return ProviderMetadata('test', '.pdf', 'text/plain', '1234', 'http://wb.osf.io/file/test.pdf?token=1234')
+    return ProviderMetadata('test', '.pdf', 'text/plain', '1234',
+                        'http://wb.osf.io/file/test.pdf?token=1234')
```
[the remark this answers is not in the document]
AddisonSchiller: complete

tests/extensions/pdf/test_exporter.py
AddisonSchiller: reworded/fixed typos/added new typos
complete

mfr/extensions/pdf/settings.py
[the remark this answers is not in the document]
AddisonSchiller: Other small note: even though the list is turned into a string, it will still find '.pdf' in that string and exclude it. Not sure if its intended to work that way, but that must be why its working for images settings.  (this is why it works even though it seems like it shouldn't)

As mentioned above im open to any fix for this

tests/extensions/pdf/test_exporter.py
AddisonSchiller: I found some libraries that use system libraries to check things (libmagic used by python-magic) etc. They would require adding things to the docker file or something like that. I just went with opening the file, and pulling the signature out directly and looking at it. Slightly crude but works fine.

AddisonSchiller:
@felliott 
Random comments/CRR

Take a look at the settings questions, not quite sure how you want me to handle that.

On libtiff:
I took a look at it, and it was sort of inconclusive. All i could find is there seem to be some tiff files that label things not to the tiff standard. using libtiff might help with those? I don't think we are using it in our docker container atm (it might come auto installed though.)

I did try to install it and see if it changed the way any of my test tiff files rendered and it did not seem to. 

Just couldn't find enough information about it. Some of the docs linked to other libraries i originally looked at for rendering tiffs etc, and such. If you want i could definitely spend more time on this.

Also, leaving in additional development while I wait for responses to some of my questions

mfr/extensions/pdf/settings.py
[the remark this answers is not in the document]
felliott: Blarf, sorry I was misremembering. The similar example I was thinking of was [here](https://github.com/CenterForOpenScience/modular-file-renderer/blob/develop/mfr/server/settings.py#L35).  We support passing a list via the env by requiring it to be a space-separated string that gets split upon construction.  And yes, the image renderer settings should be doing it this way, but we never bothered to audit the settings when we set it up.  There are probably lots of settings that need this, so we'll create a separate janitorial ticket to audit and clean these up.

tests/extensions/pdf/test_exporter.py
felliott: :+1:

tests/extensions/pdf/test_exporter.py
felliott: The three passing tests are all very similar, they just take slightly different arguments.  You could parameterize them a la [this WB s3 test](https://github.com/CenterForOpenScience/waterbutler/blob/develop/tests/providers/s3/test_provider.py#L119).  That way if you update one later, the others get updated, too.

felliott:
I've worked a bunch with .tiff files in the past.  The tiff format is more of a container format than a image compression format.  It permits lots of weird formats and extensions and its a pretty popular format in science for that reason. I'm not usually a fan of adding dependencies speculatively, but I would make an exception for tiffs.  It would be good to add a comment explaining why it's in there.

mfr/extensions/pdf/settings.py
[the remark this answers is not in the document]
AddisonSchiller: Done, and ticket made.

tests/extensions/pdf/test_exporter.py
AddisonSchiller: This is great,
completed.

AddisonSchiller:
Did some more research. It looks like libtiff5 comes default in Jessie. I read some things about the correct version of libtiff not being used. It shouldn't hurt to install python-libtiff and let pillow go crazy with it. So adding that to the PR as suggested.

AddisonSchiller:
Latest additions:
Confirmed (with @felliott ) that libtiff5-dev is working in the docker container
Added libtiff5-dev to the docker-file
Removed unused python libtiff thing

Turning on libtiff use in pillow fixed color issues on some tiffs
fixed resizing on large multipage tiffs (should now display all pages)
lots of refactoring/changes from first few versions.
Pillow is upgraded to 4.3.0 on this ticket. Standalone it will work with tiffs, but will break the image renderer. There is another ticket I have out that should be merged before this one that will fix what pillow 4.3.0 breaks in MFR.

felliott:
TerTIFFic!  Moving to RTM

PR #278, opened 2017-09-10
Corrected documentation

jonathon-love:


felliott:
Hey @jonathon-love!  You're right, but I think we'll try to expand it a little more.  When the OSF was Dockerized we made it a string of space-separated domains, b/c in the new docker-compose world, you have to whitelist both the OSF and WaterButler.  I'll file a ticket with our team to update the example here and in `CONTRIBUTING.rst` and to add a blurb about why it's necessary.

PR #282, opened 2017-10-04
[SVCS-488]`Public_file` query param and office365 renderer

AddisonSchiller:
refs: https://openscience.atlassian.net/browse/SVCS-488

## Purpose
docx rendering is very intensive on the osf. By using Office Online to render publicly available docx files, we can remove a lot of pressure from the unoconv container.

## Summary of changes. 
Added support for public renderers
Added support to parse for a `public_file` query param. This query param is optional.`public_file=1` denotes that the file is public, while `public_file=0` denotes that it is private. All other values for `public_file` cause errors to be raised.
ProviderMetadata now has an is_public value, which defaults to false.
Added the office365 renderer, which is a public renderer. 

## Testing Notes
the Office365 renderer does not use the pdf renderer like unoconv used to, so the pdfs that get made by this renderer will not display exactly the same. 

There is a README.md in the renderer with more information about testing. 

In order to test this with the OSF, an OSF side change needs to be made to make use of the new `public_file` query param.

### DIFF
--- a/mfr/core/provider.py
+++ b/mfr/core/provider.py
@@ -48,17 +48,19 @@ class BaseProvider(metaclass=abc.ABCMeta):
 
 class ProviderMetadata:
 
-    def __init__(self, name, ext, content_type, unique_key, download_url):
+    def __init__(self, name, ext, content_type, unique_key, download_url, is_public=False):
         self.name = name
         self.ext = ext
         self.content_type = content_type
         self.unique_key = unique_key
         self.download_url = download_url
+        self.is_public = is_public
 
     def serialize(self):
         return {
             'name': self.name,
             'ext': self.ext,
+            'is_public': self.is_public,
             'content_type': self.content_type,
             'unique_key': str(self.unique_key),
             'download_url': str(self.download_url),
--- a/mfr/core/utils.py
+++ b/mfr/core/utils.py
@@ -76,6 +76,18 @@ def make_renderer(name, metadata, file_path, url, assets_url, export_url):
     :rtype: :class:`mfr.core.extension.BaseRenderer`
     """
     normalized_name = (name and name.lower()) or 'none'
+    if metadata.is_public:
+        try:
+            return driver.DriverManager(
+                namespace='mfr.public_renderers',
+                name=normalized_name,
+                invoke_on_load=True,
+                invoke_args=(metadata, file_path, url, assets_url, export_url),
+            ).driver
+        except:
+            # Check for a public renderer, if one doesn't exist, use a regular one
+            pass
+
     try:
         return driver.DriverManager(
             namespace='mfr.renderers',
--- a/mfr/core/utils.py
+++ b/mfr/core/utils.py
@@ -1,7 +1,9 @@
 from stevedore import driver
 
 from mfr.core import exceptions
+import logging
 
+logger = logging.getLogger(__name__)
 
 def make_provider(name, request, url):
     """Returns an instance of :class:`mfr.core.provider.BaseProvider`
@@ -76,6 +78,7 @@ def make_renderer(name, metadata, file_path, url, assets_url, export_url):
     :rtype: :class:`mfr.core.extension.BaseRenderer`
     """
     normalized_name = (name and name.lower()) or 'none'
+    logger.info(str(metadata.is_public))
     if metadata.is_public:
         try:
             return driver.DriverManager(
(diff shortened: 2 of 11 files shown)


mfr/core/utils.py
```diff
@@ -76,6 +76,18 @@ def make_renderer(name, metadata, file_path, url, assets_url, export_url):
...
     normalized_name = (name and name.lower()) or 'none'
+    if metadata.is_public:
+        try:
+            return driver.DriverManager(
+                namespace='mfr.public_renderers',
+                name=normalized_name,
+                invoke_on_load=True,
+                invoke_args=(metadata, file_path, url, assets_url, export_url),
+            ).driver
+        except:
```
cslzchen: Be more explicit on the exception?

mfr/extensions/office365/README.md
```diff
@@ -0,0 +1,20 @@
+
+# Office 365 Renderer
```
cslzchen: 👍 

Need to update this later when OSF side is updated.

mfr/extensions/office365/render.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+
+    TEMPLATE = TemplateLookup(
+        directories=[
+            os.path.join(os.path.dirname(__file__), 'templates')
+        ]).get_template('viewer.mako')
+
+    def render(self):
+        download_url = furl.furl(self.metadata.download_url).set(query='')
+        url = settings.OFFICE_BASE_URL + download_url.url
```
cslzchen: Encode the URL since it is the value of the `src` query param. It works without encoding during local test since the URL is probably not ambiguous.

mfr/core/provider.py
```diff
@@ -48,17 +48,19 @@ def download(self):
 
 class ProviderMetadata:
 
-    def __init__(self, name, ext, content_type, unique_key, download_url):
+    def __init__(self, name, ext, content_type, unique_key, download_url, is_public=False):
```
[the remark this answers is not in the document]
AddisonSchiller: You mean like get rid of the public_file=false
? and it just being..
```
.... download_url, is_public)
```
?

mfr/core/utils.py
AddisonSchiller: I think we talked about this. The reason it isn't specific is because any specific exception will be reported below.
If this is new let me know.

mfr/core/provider.py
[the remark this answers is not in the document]
cslzchen: As discussed, keep it as kwargs.

mfr/extensions/office365/render.py
cslzchen: Can you try a revision download URL to see if it still works without encoding?

mfr/core/utils.py
```diff
@@ -76,6 +78,20 @@ def make_renderer(name, metadata, file_path, url, assets_url, export_url):
     :rtype: :class:`mfr.core.extension.BaseRenderer`
     """
     normalized_name = (name and name.lower()) or 'none'
+    logger.info(str(metadata.is_public))
```
cslzchen: - [ ] Remove this line.

Probably need to remove the imports as well.

mfr/extensions/office365/render.py
cslzchen: - [ ] Need proper encoding.

What we feed to MS Office Live is an WB URL and here is an example of my local one which is more complicated than OSF url we test.

```
http://192.168.168.167:7777/v1/resources/ygcb4/providers/osfstorage/5a01dd60005aca00414c567d?action=download&public_file=1&version=1&mode=render&direct
```

mfr/core/provider.py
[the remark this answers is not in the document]
cslzchen: As discussed, keep it as `kwargs`.

mfr/extensions/office365/render.py
AddisonSchiller: Here is what mine looks like
This is the `download_url` after the query params are stripped off
```
http://192.168.168.167:7777/v1/resources/3uq58/providers/osfstorage/5a03579db4eef2001827a8c3
```
as long as the project is public, going to that link will download the file. Also there are no chars that -need- to be encoded. If they should be encoded, well probably?  We briefly discussed this. Thoughts on encoding the `/` marks?

cslzchen:
Replaced by https://github.com/CenterForOpenScience/modular-file-renderer/pull/304.

PR #283, opened 2017-10-04
[SVCS-525] Adding documentation for supported filetypes

AddisonSchiller:
refs: https://openscience.atlassian.net/browse/SVCS-525

## Purpose
Add a public human readable list of file types supported by mfr
Add tests for this documentation

## Summary of changes
Added a new file in the top level directory with a list of supported file types
Added tests to make sure it gets updated.

## QA/Testing
Documentation change, none needed.


felliott:
Agreed, since we'll be adding a bunch of these, it would be good to have a way to autogen it.

...or....

What if we keep this as a static file, but add something to `test/` to make sure everything in the setup.py is mentioned in the doc file? Then updating to the list becomes a prerequisite of submitting a PR.  Otherwise travis will fail and yell at the submitter.  Would that be easier?

Opinions? @cslzchen, @AddisonSchiller?

cslzchen:
I like the second idea. It is visible to both `test` and CR. (I have no clue how to do it through autogen ...)

AddisonSchiller:
I could do it either way. I like the idea of travis yelling at people (probably mostly me).

For autogen, where could we set it up so it runs often enough? Or would it be a script to run before you make a PR that adds/removes new file types?

felliott:
Let's skip autogen and just add a validation test.  That also gives us more control over the formatting and we don't have to worry about forgetting to run it.  Please add that to this PR, then it LGTM.  The test can dogfood its own PR.

tests/documentation/test_entrypoints.py
```diff
@@ -0,0 +1,23 @@
...
+            readme_ext = [line.strip()[2:] for line in file if '*' in line]
+        setup_path = os.path.join(os.path.dirname((parent_dir)), 'setup.py')
+        with open(setup_path, 'r') as file:
+            parse = False
+            setup_ext = []
+            for line in file:
+                if parse and '#' not in line and ' = ' in line:
+                    setup_ext.append(line.strip().split(' = ')[0][1:])
+                # after you see this line its okay to start parsing
+                if '\'mfr.renderers\': [' in line:
```
cslzchen: - [ ] To make this more robust and more efficient, we can do equality check on the trimmed line.

tests/documentation/test_entrypoints.py
```diff
@@ -0,0 +1,23 @@
...
+        parent_dir = os.pardir
+        readme_path = os.path.join(os.path.dirname((parent_dir)), 'supportedextensions.md')
+        with open(readme_path, 'r') as file:
+            readme_ext = [line.strip()[2:] for line in file if '*' in line]
+        setup_path = os.path.join(os.path.dirname((parent_dir)), 'setup.py')
+        with open(setup_path, 'r') as file:
+            parse = False
+            setup_ext = []
+            for line in file:
+                if parse and '#' not in line and ' = ' in line:
```
cslzchen: - [ ] In case someone put a comment after the line, lets check if the trimmed line starts with `'#'`.

tests/documentation/test_entrypoints.py
cslzchen: Here is an example of a more efficient `if` statement:

```python
for line in file:
    line = line.strip()
    if parse:
        if line.startswith('#') and ' = ' in line:
            ...
    elif line == '\'mfr.renderers\': ['
        ...
```

CONTRIBUTING.rst
```diff
@@ -8,6 +8,7 @@ In general
 - `PEP 8`_, when sensible.
 - Test ruthlessly. Write docs for new features.
 - Even more important than Test-Driven Development--*Human-Driven Development*.
+- If you add an extension to setup.py, add it to supportedextensions.md.
```
cslzchen: 👍

tests/documentation/test_entrypoints.py
AddisonSchiller: Changed test to completely get rid of If statements. this makes it much cleaner and nicer.
Found out how to get access to the pkg_resources from a PR @Johnetordoff  put in (thanks John!). Had googled for it earlier but couldn't find it.

felliott:
Sweet!  Merging.

PR #284, opened 2017-10-05
[SVCS-507] Fix tabular renderer ext matching error

AddisonSchiller:
refs: https://openscience.atlassian.net/browse/SVCS-507

## Purpose
If a tabular file extension was uppercased anywhere, it would throw an error. This changes allows the library matching function to ignore casing and find the proper function to render a file.

## Summary of Changes
function_preference is now found from extension.lower() instead of just the extension

## QA/Test Notes

There are 2 files on the ticket to test this with. Before changes, they should cause an error. With this change, they should render normally.

### DIFF
--- a/mfr/extensions/tabular/render.py
+++ b/mfr/extensions/tabular/render.py
@@ -63,7 +63,7 @@ class TabularRenderer(extension.BaseRenderer):
         :param ext: file extension
         :return: tuple of column headers and row data
         """
-        function_preference = settings.LIBS.get(ext)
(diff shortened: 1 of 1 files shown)


felliott:
@cslzchen: what do you mean by "fix for the rest"?  I think (not positive) that the tabular renderer is the only one that does a subdispatch within the renderer.  The others are dispatched in the server handler, which already lowercases.

felliott:
Merged.

PR #288, opened 2017-10-12
[SVCS-530] Xlsx duplicate header error fix

AddisonSchiller:
## Ticket
https://openscience.atlassian.net/browse/SVCS-530

## Purpose
The xlsx renderer tool for the tabular renderer was overwriting column values if there were duplicate names in the header names.

## Changes
The tabular renderer will no longer overwrite the values
for headers that have the same name. Instead it will rename
all duplicated headers in the format `name (1)`
There is a very unlikely case where, if after searching for a name for 5000 iterations, it will use a UUID instead of a count.

Added some tests (One is commented out because of how hard it is to test)

## Side effects
None that I know of

## QA Notes
There is a zip file on the JIRA ticket with files to test with.


mfr/extensions/tabular/libs/xlrd_tools.py
```diff
@@ -34,6 +36,30 @@ def xlsx_xlrd(fp):
...
+                if fields[x] in duplicate_fields:
+                    name = fields[x]
+                    increased_name = name + ' ({})'.format(counts[name])
+                    # this triggers if you try to rename a header, and that new name
+                    # already exists in fields. it will then increment to look for the
+                    # next available name.
+                    iteration = 0
+                    while increased_name in fields:
+                        iteration += 1
+                        if iteration > 5000:
```
cslzchen: Set iteration cap as a default argument and use a lower number for testing.

tests/extensions/tabular/test_xlsx_tools.py
```diff
@@ -17,4 +19,34 @@ def test_xlsx_xlrd(self):
...
+        assert sheet[0][3] == {'id': 'Dup (3)', 'name': 'Dup (3)',
+                                'field': 'Dup (3)', 'sortable': True}
+        assert sheet[0][4] == {'id': 'Dup (4)', 'name': 'Dup (4)',
+                                'field': 'Dup (4)', 'sortable': True}
+        assert sheet[0][5] == {'id': 'Not Dup', 'name': 'Not Dup',
+                                'field': 'Not Dup', 'sortable': True}
+        assert sheet[1][0] == {'Name': 1.0, 'Dup (1)': 2.0, 'Dup (2)': 3.0,
+                            'Dup (3)': 4.0, 'Dup (4)': 5.0, 'Not Dup': 6.0}
+
+    # After demo it was suggested the iteration cap be raised. The value ended up to be about 5,000
```
cslzchen: As suggested above, use a default arg for `iterations` and set it lower. You can then use this for this test instead of having to make a file to iterate 5000 times.

AddisonSchiller:
@cslzchen , added max_iterations variable for testing. Re-enabled `uuid` test. Also a few minor style changes (renaming vars etc)

mfr/extensions/tabular/libs/xlrd_tools.py
```diff
@@ -1,13 +1,15 @@
...
+import xlrd
+
 from mfr.extensions.tabular.compat import range, basestring
+from mfr.extensions.tabular.utilities import header_population
+from mfr.extensions.tabular.exceptions import TableTooBigError
 
 
-def xlsx_xlrd(fp):
-    """Read and convert a xlsx file to JSON format using the xlrd library
+def xlsx_xlrd(fp, max_iterations=5000):
```
cslzchen: 👍

PR #291, opened 2017-10-24
[SVCS-565] Add sandboxing to Iframes with `allow-scripts`

AddisonSchiller:
## Ticket
https://openscience.atlassian.net/browse/SVCS-565

## Purpose
Add sandboxing to MFR Iframe

## Changes
Added sandbox with "allow-scripts" (many renders will not work if you disable javascript)

## Side effects
could possibly break a few remote untested file types, but probably not.
Could cause problems for future renderers/file types

## QA Notes
Could use a wider array of tests. I tested each renderer a few times, but not all file types (and not many edge cases. This will stop things like links in pdfs, or atleast it should I think?)

### DIFF
--- a/mfr/server/static/js/mfr.js
+++ b/mfr/server/static/js/mfr.js
@@ -82,6 +82,7 @@
             self.pymParent.iframe.setAttribute('scrolling', 'yes');
             self.pymParent.iframe.setAttribute('horizontalscrolling', 'yes');
             self.pymParent.iframe.setAttribute('verticalscrolling', 'yes');
+            self.pymParent.iframe.setAttribute('sandbox', 'allow-scripts');
 
             self.pymParent.el.appendChild(self.spinner);
             $(self.pymParent.iframe).on('load', function () {


mfr/server/static/js/mfr.js
```diff
@@ -82,6 +82,7 @@
             self.pymParent.iframe.setAttribute('scrolling', 'yes');
             self.pymParent.iframe.setAttribute('horizontalscrolling', 'yes');
             self.pymParent.iframe.setAttribute('verticalscrolling', 'yes');
+            self.pymParent.iframe.setAttribute('sandbox', 'allow-scripts');
```
cslzchen: I think it is ok for us to sandboxing with `allow-scripts`. However, we want to try it out if we can give the iframe a different domain to use the protection from Same Origin Policy (if it is easy and if it breaks nothing).

AddisonSchiller: If you don't set `allow-same-origin` it just treats all things as different origin. From my tests we do not need `allow-same-origin` for anything specific to render, so I did not include it.

Am I understanding you right?

cslzchen: I misunderstood the `sandbox` attribute. Looks good! 🎆  Let me do a local test.

felliott:
Looks good, merging.

PR #292, opened 2017-10-24
[SVCS-528] Let WB know a request is coming from MFR

AddisonSchiller:
## Ticket
https://openscience.atlassian.net/browse/SVCS-528

Related Waterbutler counterpart PR:  https://github.com/CenterForOpenScience/waterbutler/pull/290

## Changes
Append a `mfr` query param to WB urls to let it know the request is coming from WB
Move .gdoc extension from unoconv renderer/exporter to pdf renderer

## Side effects
As long as the `mfr` query param is in effect, none

## QA Notes
This must be tested with its WB counterpart to have any effect.

### DIFF
--- a/mfr/server/handlers/core.py
+++ b/mfr/server/handlers/core.py
@@ -104,6 +104,8 @@ class BaseHandler(CorsMixin, tornado.web.RequestHandler, SentryMixin):
 
         try:
             self.url = self.request.query_arguments['url'][0].decode('utf-8')
+            # TODO should probably use furl here, but maybe wait till after furl is upgraded
+            self.url += '&mfr=true'
         except KeyError:
             raise exceptions.ProviderError(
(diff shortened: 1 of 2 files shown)


mfr/server/handlers/core.py
```diff
@@ -104,6 +104,8 @@ def NAME(self):
 
         try:
             self.url = self.request.query_arguments['url'][0].decode('utf-8')
+            # TODO should probably use furl here, but maybe wait till after furl is upgraded
+            self.url += '&mfr=true'
```
cslzchen: This is fine since `self.url` is a literal string instead of an object as long as the URL does not contain fragment parameters.

cslzchen:
Replaces by https://github.com/CenterForOpenScience/modular-file-renderer/pull/303.

PR #293, opened 2017-10-25
[SVCS-566]Use CSS spinner over deprecated one

AddisonSchiller:
## Ticket
https://openscience.atlassian.net/browse/SVCS-566

## Purpose
Get rid of old loading spinner

## Changes
replaced old spinner with new pure css osf-style spinner

## Side effects
None


mfr/server/static/js/mfr.js
```diff
@@ -68,12 +67,12 @@
      * @param {Object} config Configuration to override the default settings.
      * @param {String} imgName The filename of an image in mfr/server/static/images/ to use as a loading spinner
      */
-    lib.Render = function (id, url, config, imgName) {
+    lib.Render = function (id, url, config, imgName=null) {
```
cslzchen: This doesn't look right. Can we just remove this argument `imgName`? But I am not seeing any javascript error from the browser console and rendering works as expected.

AddisonSchiller: I wasn't as clear on the jira ticket on this as i should have been. That function is called by the OSF side of things. It asks for a specific spinner ie `cos_logo.png`. The new CSS spinner does not need an `imgName`, but the argument still needs to exist or else things won't render on the OSF (will just throw an error) So i left in `imgName=null` so the OSF can still give it the arg and not break. Also once it is removed it won't break either and then `imgName` can be taken out for good.

cslzchen: Using `null` as a default value isn't the way to go though. In javascript, the default value is `undefined` (no need to explicitly set it). And since we no longer use `imgName` in the function, we can set it to `undefined` to overwrite the value that is passed into.

```javascript
lib.Render = function (id, url, config, imgName) {
    // some comment
    imgName = undefined 
}
```

AddisonSchiller: I understand now, makes sense to me. Too used to python syntax where 
`lib.Render('12', some_url, some_config)` would throw an error since there is no imgName passed in.

AddisonSchiller: Fixed

felliott:
Looks good, moving to RTM.

PR #294, opened 2017-11-01
[SVCS-556]Bump Codepygments version

AddisonSchiller:
## Ticket

https://openscience.atlassian.net/browse/SVCS-556

## Purpose
Update Codepygments version

## Changes
Updated Codepyigments python library to 2.2.0

## QA Notes
Should be tested with many different file types rendered by the Codepygments renderer. (if you need a list I can provide one upon request) 
Full regression on Codepygments renderer.
Should not effect any other aspect of MFR.


felliott:
Looks good, moving to Ready to Merge. Will go into v0.24.

PR #296, opened 2017-11-09
[SVCS-159] Add renderer/exporter names to cached files

AddisonSchiller:
## Ticket
https://openscience.atlassian.net/browse/SVCS-159

## Purpose
Add renderer name to cached files
Add exporter name to cached files

## Changes
Added renderer to cached renderer files
Added exporter name to cached exported files
Added tests for changes

## Side effects
Should be none

## QA Notes
In order to test this locally, you need to turn MFR caching on: https://github.com/CenterForOpenScience/modular-file-renderer/blob/develop/mfr/server/settings.py#L28

Once this is on, in your logs you will see the caching messages, and it will show you the names of the cached files as well. 

The tests do not directly test this change in action, instead it tests to make sure that the added `get_exporter_name` and `get_renderer_name` return the proper values.

## Deployment Notes
Some things to note: While renderer/exporter names have been added to cached files, this isn't always indicative of which files should be cleared when changes are made to a specific renderer.

For instance a cached .docx file will have `.UnoconvRenderer` at the end of it, but it is rendered by both the `.UnoconvRenderer` and the `PdfRenderer`.

### DIFF
--- a/mfr/core/utils.py
+++ b/mfr/core/utils.py
@@ -1,3 +1,4 @@
+import pkg_resources
 from stevedore import driver
 
 from mfr.core import exceptions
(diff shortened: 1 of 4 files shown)


mfr/server/handlers/export.py
```diff
@@ -29,19 +29,20 @@ class ExportHandler(core.BaseHandler):
...
+            '/export/{}.{}'.format(self.cache_file_id, self.exporter_name)
         )
         self.source_file_path = await self.local_cache_provider.validate_path(
             '/export/{}'.format(self.source_file_id)
         )
 
         self.output_file_id = '{}.{}'.format(self.source_file_path.name, self.format)
         self.output_file_path = await self.local_cache_provider.validate_path(
-            '/export/{}'.format(self.output_file_id)
+            '/export/{}.{}'.format(self.output_file_id, self.exporter_name)
```
cslzchen: I remember we did want to change something here during discussion but I forget the detail. My bad ... :-(.

cslzchen:
Really like the comments and doc str, very helpful 🎆 🎆

mfr/server/handlers/export.py
AddisonSchiller: I think i remember, 👍

mfr/core/utils.py
```diff
@@ -97,6 +98,46 @@ def make_renderer(name, metadata, file_path, url, assets_url, export_url):
...
+    """
+
+    # `make_renderer` should have already caught if an extension doesn't exist.
+
+    # should be a list of length one, since we don't have multiple entrypoints per group
+    entry_attrs = pkg_resources.iter_entry_points(group='mfr.exporters', name=name)
+
+    # ep.attrs is a tuple of attributes. There should only ever be one or `None`.
+    # For our case however there shouldn't be `None`
+    return list(entry_attrs)[0].attrs[0]
```
cslzchen: `list(entry_attrs)[0].attrs[0]` always breaks with index error.

cslzchen: `list(entry_attrs)[0].attrs[0]` always fails with index error.

cslzchen: The original code works. However, `entry_attrs` is iterator but accessed in an `array` way. The first call to `list(entry_attrs)` clears moves the iteration foward, thus further access will fail with index error. This will be updated in https://github.com/CenterForOpenScience/modular-file-renderer/pull/320.

PR #298, opened 2017-11-10
[SVCS-577] Scrub periods out of keen payload keys

AddisonSchiller:
## Ticket
https://openscience.atlassian.net/browse/SVCS-577

## Purpose
Scrub '.' from keys in keen payloads. 

## Changes
'.' are now removed from keen payloads. If this created a conflict with another key, a (1) will be added to the end of it. Increments based on number of collisions
Added tests for new scrubbing function

## Side effects
Should be none

## QA Notes
To test this locally, you need to make a keen account at https://keen.io/ and then add your keys to
the settings file.

Verify that your key is working by browsing a file or downloading a file on the osf. You should get some data to show up on your Keen.

Next, you need to send a bad payload. The best way to do this is to manually add code to the `_send_to_keen()` funtction.
```python
payload[test.....in....name] = 'dict value'
```

Once logged to keen, the value should appear as `testinname = 'dict value'`
Name incrementation can be tested like this:

```python
payload['test....key] = 'some value'
payload[test.key] = 'other value'
```
once scrubbed, this will create a conflict. it will be solved by one of the keys turning into a new key

```python
'testkey' : .... ,
'testkey (1)` value, ...
```

Unsure of best way to test this on staging, or how to send a dummy payload there.


felliott:
Delightful! Moving to RTM.

PR #300, opened 2017-11-14
[SVCS-565]Allow same origin in sandboxing

AddisonSchiller:
## Ticket
https://openscience.atlassian.net/browse/SVCS-565

## Purpose
Add `allow-same-origin` to our sandboxing

## Changes
added `allow-same-origin` to sandboxing. this should fix the PDF issue on staging (hopefully)

## Side effects
Could possibly make it less secure (specifically jasp and jamovi renderers)


felliott:
Merged!

PR #305, opened 2017-12-07
[SVCS-418] [WIP] Improve md rendering

cslzchen:
### Ticket

https://openscience.atlassian.net/browse/SVCS-418
This PR replaces: https://github.com/CenterForOpenScience/modular-file-renderer/pull/272

### Purpose

TBD

### Changes

TBD

### Side Effects

TBD

### QA Notes

TBD

### Deployment Notes

TBD


mfr/extensions/md/static/js/markdown-it-sanitizer.min.js
```diff
@@ -0,0 +1,2 @@
+/*! markdown-it-sanitizer 0.4.3 https://github.com/svbergerem/markdown-it-sanitizer @license MIT */
```
cslzchen: 👍

mfr/extensions/md/static/js/markdown-it.min.js
```diff
@@ -0,0 +1 @@
...",lAarr:"\u21da",Lacute:"\u0139",lacute:"\u013a",laemptyv:"\u29b4",lagran:"\u2112",Lambda:"\u039b",lambda:"\u03bb",lang:"\u27e8",Lang:"\u27ea",langd:"\u2991",langle:"\u27e8",lap:"\u2a85",Laplacetrf:"\u2112",laquo:"\xab",larrb:"\u21e4",larrbfs:"\u291f",larr:"\u2190",Larr:"\u219e",lArr:"\u21d0",larrfs:"\u291d",larrhk:"\u21a9",larrlp:"\u21ab",larrpl:"\u2939",larrsim:"\u2973",larrtl:"\u21a2",latail:"\u2919",lAtail:"\u291b",lat:"\u2aab",late:"\u2aad",lates:"\u2aad\ufe00",lbarr:"\u290c",lBarr:"\u290e",lbbrk:"\u2772",lbrace:"{",lbrack:"[",lbrke:"\u298b",lbrksld:"\u298f",lbrkslu:"\u298d",Lcaron:"\u013d",lcaron:"\u013e",Lcedil:"\u013b",lcedil:"\u013c",lceil:"\u2308",lcub:"{",Lcy:"\u041b",lcy:"\u043b",ldca:"\u2936",ldquo:"\u201c",ldquor:"\u201e",ldrdhar:"\u2967",ldrushar:"\u294b",ldsh:"\u21b2",le:"\u2264",lE:"\u2266",LeftAngleBracket:"\u27e8",LeftArrowBar:"\u21e4",leftarrow:"\u2190",LeftArrow:"\u2190",Leftarrow:"\u21d0",LeftArrowRightArrow:"\u21c6",leftarrowtail:"\u21a2",LeftCeiling:"\u2308",LeftDoubleBracket:"\u27e6",LeftDownTeeVector:"\u2961",LeftDownVectorBar:"\u2959",LeftDownVector:"\u21c3",LeftFloor:"\u230a",leftharpoondown:"\u21bd",leftharpoonup:"\u21bc",leftleftarrows:"\u21c7",leftrig
```
cslzchen: Where does this come from? Which version? Please add a comment at the beginning of the file.

tests/extensions/md/test_renderer.py
```diff
@@ -48,38 +50,25 @@ def test_render_md_cache_result(self, renderer):
...
+            '1234',
+            'http://wb.osf.io/file/test.md?token=1234'
+        )
+
+        md_renderer = MdRenderer(metadata, test_md_file_path, url, assets_url, export_url)
+        body = md_renderer.render()
+        in_body = 'The rain---not the reign---in\nSpain.\n\n&lt;script&gt;\nalert("Hello world");\n&lt;/script&gt;'
+        assert in_body in body
+
+    # TODO: it is hard to test since rendering is performed by front-end javascript, if possible test the following:
```
cslzchen: It is hard to test front-end based rending for `.md` and there a lot to test. I will try to add more tests before/during phase 2 review if we think it is necessary.

.eslintrc.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
```
cslzchen: Add a basic configuration for `eslint`.

.gitignore
```diff
@@ -11,19 +11,29 @@ Thumbs.db
 *.swp
 *~
 
+# Node Modules
+######################
+/node_modules/
```
cslzchen: Update `.gitignore` to exclude `./node_modules/` (here) and `./src/` (below).

mfr/extensions/md/render.py
```diff
@@ -23,12 +16,11 @@ class MdRenderer(extension.BaseRenderer):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.metrics.add('markdown_version', markdown.version)
 
     def render(self):
         """Render a markdown file to html."""
         with open(self.file_path, 'r') as fp:
-            body = markdown.markdown(fp.read(), extensions=[EscapeHtml()])
+            body = bleach.clean(fp.read(), **md_settings.BLEACH_WHITELIST)
```
cslzchen: Although MFR uses `bleach` the same way OSF does, the library is not exactly the same due to python version difference.

AddisonSchiller: Not sure if it got noted somewhere else, but more specifically it causes problems with unclosed tags.

```
Mark down is great!
<p> okay </p

other stuff
more stuff

```
will usually just cut off the last 2 lines and end at at the 'okay' in a completed `<p>` tag

cslzchen: I see. My test didn't have the extra lines which made me believe that python35 closes unclosed tags correctly. I will test this again. Thanks for the follow up @AddisonSchiller

AddisonSchiller: I can't remember the exact behavior, but I do recall it working correctly sometimes with unclosed tags, and other times displaying the above behavior. May need a bit more investigation to figure out the full extent of it. Bleach also had some github issues that explained the behavior as well. They may be linked on an old comment somewhere, or on the old PR. May be worth checking those out.

tests/extensions/md/test_renderer.py
AddisonSchiller: Not sure how relevant this is, but I mostly just tried to test bleach quickly here. It may be worth while to make sure proper tags are being bleached, and javascript would get escaped out etc. The above is already testing for that a little.
Might be worth a parameterized test to test all the `BLEACH_WHITELIST` tags or whatever that setting got called.

cslzchen:
@felliott Thanks for the review. Need to take another look since it's been a while. Hopefully later this week.

mfr/extensions/md/render.py
```diff
@@ -23,13 +14,10 @@ class MdRenderer(extension.BaseRenderer):
...
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.metrics.add('markdown_version', markdown.version)
 
     def render(self):
         """Render a markdown file to html."""
-        with open(self.file_path, 'r') as fp:
-            body = markdown.markdown(fp.read(), extensions=[EscapeHtml()])
-            return self.TEMPLATE.render(base=self.assets_url, body=body)
+        return self.TEMPLATE.render(base=self.assets_url, url=self.metadata.download_url)
```
cslzchen: MD renderer is now of type **direct-to-wb** instead of **through-renderer**. Only pass the WB download URL to the template.

mfr/extensions/md/templates/viewer.mako
```diff
@@ -1,8 +1,74 @@
...
+<script src="${base}/js/markdown-it-mathjax.js"></script>
+<script src="${base}/js/markdown-it-sanitizer.min.js"></script>
+<script src="${base}/js/markdown-it-highlightjs.js"></script>
+<script src="${base}/js/markdown-it-ins-del.js"></script>
+<script src="${base}/js/markdown-it-toc.js"></script>
+
+<script>
+
+    ## How to load ``markdown-it``: https://github.com/markdown-it/markdown-it#simple
+    var MarkdownIt = window.markdownit;
```
cslzchen: Use `window.<LIBRARIES>` to load the libraries which have already been wrapped with `root.<LIBRARIES>` due to  MFR lacking a package manager.

mfr/extensions/md/templates/viewer.mako
```diff
@@ -1,8 +1,74 @@
...
+
+    var wb_request = new XMLHttpRequest();
+    var wb_download_url = "${url}";
+    wb_request.open("GET", wb_download_url);
+    wb_request.responseType = "text";
+    wb_request.withCredentials = true;
+    wb_request.onload = function () {
+        document.getElementById("mdViewer").innerHTML = markdown.render(wb_request.response);
+        ## Force the host to resize
+        window.pymChild.sendHeight();
```
cslzchen: iframe has a min height of 150 pixels, which is the height before its innerHTML is updated. Without this line, it doesn't resize to proper height unless there is a certain user action (scrolling, clicking, resizing, etc.) on the page.

tests/extensions/md/test_renderer.py
```diff
@@ -35,51 +32,18 @@ def export_url():
...
-        assert inbody in body
+    def test_render_md_file_required(self, mock_renderer):
+        assert mock_renderer.file_required is True
+
+    def test_render_md_cache_result(self, mock_renderer):
+        assert mock_renderer.cache_result is True
 
+    def test_render_md(self, mock_renderer):
+        body = mock_renderer.render()
+        assert mock_renderer.metadata.download_url in body
```
cslzchen: The only thing that needs testing is that the download URL are correctly passed to and parsed by the template.

PR #310, opened 2018-01-25
Fix to rendering of .omv files with images in them

jonathon-love:
## Purpose

Fixes a bug in the rendering of .omv files which contain images. (Image urls containing `%20` were not decoded back to spaces, and reading images from the archive was failing.)

## Changes

Image urls are now decoded correctly

## Side effects

None

## QA Notes

An additional test has been added to the jamovi tests

## Deployment Notes

Nope


felliott:
Merged and verified on our testing environment.  I'll update production once travis finishes.  Thanks, @jonathon-love!

felliott:
Deployed!

jonathon-love:
thanks @felliott! i was worried it was going to take 4 months to be deployed, so to have a turn around in less than 24 hours is super appreciated!

jonathon

PR #312, opened 2018-02-07
[] [] Don't try to remove the temporary dir on the first install

birdbrained:
## Purpose

Running `docker-compose up mfr_requirements` fails because it waits for user interaction. Removing `-e` lets pip install aiohttppretty without failing because pip doesn't know what to do.

## Changes

Remove the `-e` flag from the requirements file

## Side effects

This flag looks like its been here since the dependency was first added, but there is no reason stated. There may be a reason its there? If so we can look at `--ignore-installed` as an alternative

## QA Notes

Probably no.

## Deployment Notes

Probably no. This only affects dev environments.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,6 +1,6 @@
 -r requirements.txt
 
--e git+https://github.com/centerforopenscience/aiohttpretty.git@0.0.2#egg=aiohttpretty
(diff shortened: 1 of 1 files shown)


birdbrained:
My bad, I thought my newline was putting it in the body not the subject. (Normally I don't use `commit -m` but an editor to write commit messages. Serves me right for trying something new.)

felliott:
LGTM!  Merging.

PR #318, opened 2018-03-19
[SVCS-671] Fix scrollbars on video

birdbrained:
## Ticket

https://openscience.atlassian.net/browse/SVCS-671
## Purpose

This makes it so videos rendered by the MFR will not display with scrollbars inside the OSF.

## Changes

Removes any space between the edges of the iframes interior and the video.

## Side effects
doubtful, though it was mentioned sometimes wierd stuff happens in IE.

### DIFF
--- a/mfr/extensions/video/templates/viewer.mako
+++ b/mfr/extensions/video/templates/viewer.mako
@@ -1,3 +1,10 @@
+<style>
+    body {
+        margin: 0;
+        padding: 0;
+    }
+</style>
+
 <video controls height="100%">
   <source src="${url}">
   Your browser does not support the video tag.


mfr/extensions/video/templates/viewer.mako
```diff
@@ -7,4 +14,4 @@
 <script src="/static/js/mfr.child.js"></script>
 <script>
     window.pymChild.sendMessage('embed', 'embed-responsive-16by9');
-</script>
\ No newline at end of file
+</script>
```
cslzchen: 👍 for an empty line for GitHub

mfr/extensions/video/templates/viewer.mako
```diff
@@ -1,3 +1,10 @@
+<style>
```
cslzchen: * Can you add a `mako` comment `##` explaining why this works? If it is purely magic, then what it does/fixes is fine as well. In this way, others can know why this is a must-have piece without look a the commit message.

* I had worried that `<style>` may not work in `body` but local testing confirmed that MFR put it in the `<header>` tag. 😌 

video-style

birdbrained: Its not magic - its just default browser styles add margin and padding on the body tag - I thought this was common knowledge but I can add a comment if we need it. The only reason this is in a style tag and not in a CSS file is because this way we don't need to worry about it affecting other extensions by accident - this will only affect videos. Lastly yep style tags will work anywhere. html5 spec actually doesn't even require head and body tags to exist - the browser automatically inserts them if they're not there.

mfr/extensions/video/templates/viewer.mako
birdbrained: It wasn't me it was my vim settings.

mfr/extensions/video/templates/viewer.mako
cslzchen: - I see. Thanks for the clarification. I didn't understand why this solve the scroll bar flickering issue and was asking that a comment is helpful here. I tried to use this solution to fix similar scroll bar flickering issue on chrome for images but it didn't work.
- We can not rely on HTML5 since there is IE11. This is where my initial worry came from until I successfully tested this in IE11.

PR #321, opened 2018-03-21
[SVCS-674][PLAT-218] Relay MFR header requests to WB

Johnetordoff:
## Ticket
[SVCS-674](https://openscience.atlassian.net/browse/SVCS-674)
https://openscience.atlassian.net/browse/PLAT-709

## Purpose

Currently the MFR doesn't pass any useful information about the render back to the osf. This fix in conjunction with https://github.com/CenterForOpenScience/waterbutler/pull/324 will allow osf.io to receive information more information about how a file was received for analytics purposes.

## Changes

Relays MFRs Request headers to WB.

## Side effects

None that I know of.

### DIFF
--- a/mfr/providers/osf/provider.py
+++ b/mfr/providers/osf/provider.py
@@ -125,6 +125,7 @@ class OsfProvider(provider.BaseProvider):
         """Download file from WaterButler, returning stream."""
         download_url = await self._fetch_download_url()
         headers = {settings.MFR_IDENTIFYING_HEADER: '1'}
+        headers.update(self.request.headers)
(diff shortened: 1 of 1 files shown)


birdbrained:
This seems fine to me. Only thing to be aware of is that this happens - would it be worth putting in documentation that any requests to mfr forward all headers to the third parties where the file is hosted?

felliott:
After an in-person discussion with @Johnetordoff, we've decided to close this PR and continue the work in a new one.  Instead of relaying all the headers/request metadata, we'll relay the important ones (referrer and request url) in specially named headers.

PR #323, opened 2018-03-26
Update .gitignore

birdbrained:
## Purpose

Don't add node_modules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,5 @@ player/static/mfr/*
 !.gitkeep
 
 .cache
+
+node_modules/


.gitignore
```diff
@@ -72,3 +72,5 @@ player/static/mfr/*
 !.gitkeep
 
 .cache
+
+node_modules/
```
cslzchen: `/node_modules/` is preferred if  you target the top level directory. `node_modules/` ignores every directory in the whole tree if match.

felliott: Hold off please @cslzchen. I'm mid-release.

PR #326, opened 2018-03-29
[HotFix] [SVCS-676] Fix Memory Leak During CSV Rendering

cslzchen:
* This doesn't solve the memory leak when `CSV` are correctly rendered. Need more investigation.
* This does significantly reduce the memory leak when `TableTooBigError` exception is thrown.

### DIFF
--- a/mfr/extensions/tabular/render.py
+++ b/mfr/extensions/tabular/render.py
@@ -1,4 +1,5 @@
 import json
+import gc
 import logging
 import os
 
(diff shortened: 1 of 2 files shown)


mfr/extensions/tabular/render.py
```diff
@@ -30,14 +31,23 @@ def render(self):
...
+            if sheets and size:
+                return self.TEMPLATE.render(
+                    base=self.assets_url,
+                    width=settings.TABLE_WIDTH,
+                    height=settings.TABLE_HEIGHT,
+                    sheets=json.dumps(sheets),
+                    options=json.dumps(size),
+                )
+        # gc.collect()
+        raise exceptions.TableTooBigError(
```
cslzchen: Raise exception after `with open()` has finished.

mfr/extensions/tabular/render.py
```diff
@@ -30,14 +31,23 @@ def render(self):
...
+            # gc.collect()
+            if sheets and size:
+                return self.TEMPLATE.render(
+                    base=self.assets_url,
+                    width=settings.TABLE_WIDTH,
+                    height=settings.TABLE_HEIGHT,
+                    sheets=json.dumps(sheets),
+                    options=json.dumps(size),
+                )
+        # gc.collect()
```
cslzchen: Forcing garbage collection does not help, which is a good thing. I will remove the import and these statement after CR.

mfr/extensions/tabular/render.py
```diff
@@ -74,10 +89,13 @@ def _render_grid(self, fp, ext, *args, **kwargs):
 
             nbr_rows = len(sheet[1])
             if nbr_cols > settings.MAX_SIZE or nbr_rows > settings.MAX_SIZE:
-                raise exceptions.TableTooBigError('Table is too large to render.', extension=ext,
-                                                  nbr_cols=nbr_cols, nbr_rows=nbr_rows)
+                table_too_big = True
+                break
+
+        if table_too_big:
+            return None, None, nbr_rows, nbr_cols
```
cslzchen: This is an ugly way of using python's flexibility. Please advise!

PR #328, opened 2018-04-02
[HotFix] [SVCS-???] Fix Unsupported Export Type

cslzchen:
## Ticket

TBD

- [ ] @felliott Please update the ticket and related platform one, thanks!

## Purpose

Fixed the bug where unsupported export types throw 400 assertion error instead of 400 exporter error. Remove the misleading comment which wrongly claims that `get_exporter_name()` is always called with a supported exporter. Update tests accordingly.

### Before

before

### After

after

## Side effects

No

## QA Notes

TBD

## Deployment Notes

No


mfr/core/utils.py
```diff
@@ -135,7 +134,11 @@ def get_exporter_name(name: str) -> str:
     ep_iterator = pkg_resources.iter_entry_points(group='mfr.exporters', name=name.lower())
     ep_list = list(ep_iterator)
 
-    # `make_renderer()` is called before `make_exporter()` to ensure the file type is supported
+    # Empty list indicates unsupported export type.  Return '' and let `make_exporter()` handle it.
+    if len(ep_list) == 0:
```
cslzchen: The fix follows exactly the same code used in `get_renderer_name()`. In addition, the comment is corrected.

cslzchen:
@felliott I updated both cases for `render.py` and `exporter.py`.  Ready for CR again.

PR #330, opened 2018-04-16
[][SVCS-682] MFR can't render pdfs when page title has a single quote

birdbrained:
## Ticket

https://openscience.atlassian.net/browse/SVCS-682

## Purpose

Don't fail when trying to render .docx

## Changes

Runs a regex to replace "'" in urls with "\'"

## QA Notes

Make sure .docx that have quotes in the name can load.

### DIFF
--- a/mfr/extensions/audio/render.py
+++ b/mfr/extensions/audio/render.py
@@ -3,6 +3,7 @@ import os
 from mako.lookup import TemplateLookup
 
 from mfr.core import extension
+from mfr.extensions.utils import escape_url_for_template
 
 
 class AudioRenderer(extension.BaseRenderer):
@@ -13,7 +14,8 @@ class AudioRenderer(extension.BaseRenderer):
         ]).get_template('viewer.mako')
 
     def render(self):
-        return self.TEMPLATE.render(base=self.assets_url, url=self.url)
+        safe_url = escape_url_for_template(self.url)
+        return self.TEMPLATE.render(base=self.assets_url, url=safe_url)
 
     @property
     def file_required(self):
--- a/mfr/extensions/pdf/render.py
+++ b/mfr/extensions/pdf/render.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import re
 
 import furl
 from mako.lookup import TemplateLookup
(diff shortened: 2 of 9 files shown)


setup.cfg
```diff
@@ -8,4 +8,4 @@
 [flake8]
 ignore = E501,E127,E128,E265,E301,E302,F403,E731
 max-line-length = 100
-exclude = .ropeproject,tests/*,src/*,env,venv
+exclude = .ropeproject,tests/*,src/*,env,venv,./node_modules/*
```
cslzchen: The PR updating the [flake8] has already been merged [here](https://github.com/CenterForOpenScience/modular-file-renderer/pull/329/files). Please remove this change or fix conflicts.

tasks.py
```diff
@@ -51,7 +51,7 @@ def flake(ctx):
 @task
 def test(ctx, verbose=False):
     flake(ctx)
-    cmd = 'py.test --cov-report term-missing --cov mfr tests'
+    cmd = 'py.test -s --cov-report term-missing --cov mfr tests'
```
cslzchen: Why change this? Should be a separate PR if necessary.

click
```diff
@@ -0,0 +1 @@
+tell application "System Events" to click at {880, 511}
```
cslzchen: What's this?

mfr/extensions/pdf/render.py
```diff
@@ -25,7 +26,7 @@ def render(self):
             logger.debug('Extension not found in supported list!')
             return self.TEMPLATE.render(
                 base=self.assets_url,
-                url=download_url.geturl(),
+                url=re.sub(r'\'', '\\\'', download_url.geturl()),
```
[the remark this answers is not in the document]
birdbrained: This is fine, because the quotation mark is the only charachter we need to change - It's very specifically only single quotation marks that need to be represented the way they should be represented in a single quote string. (so it doesn't conflict with the js string's quote that this is printed inside of.) If a url has `\` in it, it's not a valid url (https://tools.ietf.org/html/rfc3986#section-2)

birdbrained:
Also just testing with a docx is fine, because it would be the same issue with any other file, it's only the name of the file that we care about

mfr/extensions/pdf/render.py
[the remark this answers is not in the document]
birdbrained: Use String.raw``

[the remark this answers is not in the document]
birdbrained: Note this is a cheap messy fix

mfr/extensions/audio/render.py
```diff
@@ -13,7 +14,8 @@ class AudioRenderer(extension.BaseRenderer):
         ]).get_template('viewer.mako')
 
     def render(self):
-        return self.TEMPLATE.render(base=self.assets_url, url=self.url)
+        safe_url = escape_url_for_template(self.url)
+        return self.TEMPLATE.render(base=self.assets_url, url=safe_url)
```
cslzchen: Same fix for all eligible extensions, just different URLs. I can change the style to
```python
self.TEMPLATE.render(base=self.assets_url, url=escape_url_for_template(...))
``` 
for this and other extensions if you prefer.

PR #299, opened 2017-11-10
[SVCS-580] Collapsible directories for zip rendering

Johnetordoff:
## Ticket

https://openscience.atlassian.net/browse/SVCS-580

## Purpose
Our zip render currently lists files of all directories in one big table, other implementations have a neater collapsible hierarchical structure. This PR creates a collapsible directory UI.

## Changes

Uses two new JS libraries, jstree and jstree-table and method for the renderer that  reconstructs a file tree from a list of paths.

## Side effects

None that I know of

## QA Notes

Upload a complex zip file and click through it contents.

## Deployment Notes

None that I know of.

PR #331, opened 2018-04-16
[SVCS-580] Collapsible Zip Rendering

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/SVCS-580

## Purpose

Replaces: https://github.com/CenterForOpenScience/modular-file-renderer/pull/299

Improve our zip rendering by adding a neater collapsible hierarchical structure.

## Changes

Uses two new JavaScript libraries: [jstree](https://www.jstree.com/) and [jstree-table](https://github.com/adamjimenez/jstree-table). The core algorithm is building the tree from a list of paths.

### CR Notes

Most of the files are assets. Please focus on `mfr/extensions/zip/render.py` and `mfr/extensions/zip/templates/viewer.mako` for core functionality. The tests and fixtures also helps.

All image assets are optimized so they took 15% less than the original PR. No changes to any scripts and css files from the new jstree and jstree-table library. On exception is the style change for `mfr/extensions/zip/static/js/jstreetable.js` which was "eslint"ed due to min version not available.

## Side effects

No

## QA Notes

Upload a complex zip file and click through it contents. You may discover some hidden system files depending on what OS and what `zip` app you use which is expected.

## Deployment Notes

TBD

### DIFF
--- a/mfr/extensions/zip/render.py
+++ b/mfr/extensions/zip/render.py
@@ -52,7 +52,6 @@ class ZipRenderer(BaseRenderer):
         # Iterate through each path and build the tree
         for obj in sorted_obj_list:
             path_from_root = obj.filename
-            print(path_from_root)
             path_segments = [segment for segment in path_from_root.split('/') if segment]
             # Find the parent node of the current object, always start from the root node
             parent = tree_root
--- a/tests/extensions/zip/fixtures/obj_tree.json
+++ b/tests/extensions/zip/fixtures/obj_tree.json
@@ -149,6 +149,156 @@
             ]
         }
     ],
+    "test_file_tree_sorted": [
+        {
+            "icon": "http://mfr.osf.io/assets/zip/img/file-ext-zip.png",
+            "text": "test.zip",
+            "children": [
+                {
+                    "icon": "http://mfr.osf.io/assets/zip/img/folder.png",
+                    "text": "zip-test",
+                    "data": {
+                        "size": "",
+                        "date": "2018-04-16 11:52:50"
+                    },
+                    "children": [
+                        {
+                            "icon": "http://mfr.osf.io/assets/zip/img/file-ext-generic.png",
+                            "text": "file-generic-ext.mfr",
+                            "data": {
+                                "size": " 29B",
+                                "date": "2018-03-27 15:42:32"
+                            },
+                            "children": []
+                        },
+                        {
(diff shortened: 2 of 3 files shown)


tests/extensions/zip/fixtures/obj_list.json
```diff
@@ -0,0 +1,22 @@
+{
```
cslzchen: Here is an example of the `filename` for each `file` in `ZipFile().filelist`. It is confusing since the `filename` is the full path for a file or a folder. In the code in `render.py`, I try to make this less confusing by using `obj` for var names.

tests/extensions/zip/fixtures/obj_tree.json
```diff
@@ -0,0 +1,169 @@
+{
+    "test_file_tree": [
```
cslzchen: Here is an example of a fully built `obj_tree`. Due to the requirement of the library we use, the tree is a list which contains only the root node: `[root,]`. Each node in the tree is a dictionary with four keys: `icon`, `text`, `data` and `children`. `text` is the file/folder's name and `children` is a list of child nodes.

mfr/extensions/zip/render.py
```diff
@@ -1,36 +1,188 @@
...
+
+        zip_file = ZipFile(self.file_path, 'r')
+
+        # ``ZipFile.filelist`` contains both files and folder.  Using ``obj`` for better clarity.
+        obj_list = self.sanitize_obj_list(zip_file.filelist)
+        obj_tree = self.obj_list_to_tree(obj_list)
+
+        return self.TEMPLATE.render(data=obj_tree, base=self.assets_url)
+
+    def obj_list_to_tree(self, obj_list: list) -> List[dict]:
```
cslzchen: Please take a look at the comments in the test fixtures for the structure of `obj_list` and `obj_tree`. Hopefully the in-code comments in this method provides enough information.

I'd like to mention one quirk in the tree building process. Given an object list of ['A/ab.png', 'A/', ], the algorithm first sees object 'A/ab.png'. It tries to build the tree by adding node `A/` to the root's children list and then adding `ab.png` to `A`'s children list. However, at this time we don't have details (`icon`, `data`) of `A/`. The solution is to build the structure only in this iteration and update the details in later iterations when  `A/` is the object.

birdbrained: Add an axample of fixture to the docstring?

PR #333, opened 2018-04-19
**Blocking SVCS-699** [SVCS-678] Refactor and optimize handlers/extensions

birdbrained:
# What

Improve architecture of handlers and plugins.

- Plugins receive file stream and metadata as input, and handle what local files are needed internally
- Plugins run asynchronously, and if they run a subprocess, use asyncio.create_subprocess_exec, so that threads aren't needed.
- If a plugin has a long running process that can only be designed synchronously, a thread is created to handle only that case
- Resolving plugin names is generalized, so that only one function is needed to get the name of a plugin
- Any values needed by a handler are calculated as needed by a `@property`, so we don't waste cpu time calculating values we won't use. (This should make cache hits faster)
- The render handler handles the conversion of a requested file render so that file conversion need not occur inside a handler
- When a file is rendered, the final renderer (that renders the file post-conversion) is the only renderer loaded)
- If a file can be rendered by only generating an export url (it doesn't need the actual file), the export plugin is not loaded at all during the request by deferring the export to happen when the file stream is awaited inside of the renderer.
    - The renderer knows if it need the file, if it awaits the file, the conversion happens
    - If the renderer never awaits the file, the exporter is not loaded or run, and the conversion does not happen

#Why

A no-op is a simple operation to support, why not do it?

# Acceptance criteria

A PR to support

# QA notes

Devs will test this. Dev, please add a list of affected and unaffected file types.

Build an MFR export url requesting to export a file of type foo as type foo. This should start a download of the raw file without modification. Exception: image files should respect scaling parameters.


mfr/core/extension.py
```diff
@@ -15,38 +29,90 @@ def __init__(self, ext, source_file_path, output_file_path, format):
...
+        self.source_wb_path = await self.local_cache_provider.validate_path(
+            '/export/{}'.format(self.source_file_id)
+        )
+        self.source_file_path = self.source_wb_path.full_path
+        self.output_file_id = '{}.{}'.format(self.source_wb_path.name, self.format)
+        self.output_wb_path = await self.local_cache_provider.validate_path(
+            '/export/{}'.format(self.output_file_id)
+        )
+        self.output_file_path = self.output_wb_path.full_path
         self.exporter_metrics.merge({
```
birdbrained: I need to know more about what these metrics are trying to cover so I make sure this does'nt break data we've already collected.

mfr/core/extension.py
```diff
@@ -15,38 +29,90 @@ def __init__(self, ext, source_file_path, output_file_path, format):
...
+                'id': self.output_file_id,
+                'path': str(self.output_wb_path),
+                'provider': self.local_cache_provider.NAME,
+            }
+        })
+
+        # Execute the extension's export method asynchronously and wait for it
+        # to finish
+        loop = asyncio.get_event_loop()
+        await loop.run_in_executor(None, self.export)
```
birdbrained: Make the export object handle its own async issues.

mfr/core/extension.py
```diff
@@ -15,38 +29,90 @@ def __init__(self, ext, source_file_path, output_file_path, format):
...
+
+        # Return a stream of the converted file
+        return await self.write_to_stream()
+
+    def __del__(self):
+        self.output_fp.close()
```
birdbrained: This I need a better solution for - this only works because the exporter goes out of scope when the request finishes. It works and will be reliable, but I'd like something a little more pretty.

mfr/core/extension.py
```diff
@@ -15,38 +29,90 @@ def __init__(self, ext, source_file_path, output_file_path, format):
...
+        else:
+            cache_file_path_str = '/export/{}'.format(self.cache_file_id)
+
         self.exporter_metrics = MetricsRecord('exporter')
         if self._get_module_name():
             self.metrics = self.exporter_metrics.new_subrecord(self._get_module_name())
 
+    async def __call__(self):
+
+        self.source_wb_path = await self.local_cache_provider.validate_path(
```
birdbrained: I think maybe I'll move this path/file/stream setup stuff to its own method

mfr/core/extension.py
```diff
@@ -72,6 +138,27 @@ def __init__(self, metadata, file_path, url, assets_url, export_url):
         except AttributeError:
             pass
 
+    async def __call__(self):
+
+        self.renderer_metrics.add('class', self._get_module_name())
```
birdbrained: again I need some input on where metrics need to go.

mfr/core/extension.py
```diff
@@ -72,6 +138,27 @@ def __init__(self, metadata, file_path, url, assets_url, export_url):
...
+            )
+            await self.local_cache_provider.upload(
+                await self.provider.download(),
+                self.source_file_path
+            )
+        else:
+            self.metrics.add('source_file.upload.required', False)
+
+        loop = asyncio.get_event_loop()
+        rendition = await loop.run_in_executor(None, self.render)
```
birdbrained: make the renderer handle its own business

mfr/core/utils.py
```diff
@@ -63,6 +63,51 @@ def make_exporter(name, source_file_path, output_file_path, format):
             }
         )
 
+def bind_render(metadata, file_stream, url, assets_url, export_url):
```
birdbrained: Note: need to remove the old make_renderer/make_exporter utils in favor of the ones that treat export and render as functions.

mfr/server/handlers/core.py
```diff
@@ -86,10 +86,22 @@ def __init__(self, *args, **kwargs):
...
         self.handler_metrics.add('cache_file.result', None)
         self.handler_metrics.add('source_file.upload.required', True)
-
         self.metrics = self.handler_metrics.new_subrecord(self.NAME)
 
         self.extension_metrics = MetricsRecord('extension')
+        self._cache_provider = None
+
+    @property
+    def cache_provider(self):
```
birdbrained: make the cache provider lazy so its only instatiated if used.

mfr/server/handlers/export.py
```diff
@@ -37,61 +37,49 @@ class ExportHandler(core.BaseHandler):
...
+        """Export a file to the format specified via the associated extension
+        library.
 
+        First, try to export the file directly - this is only valid if the file is
+        already in the correct format.
+
+        Next, if caching is enabled, try to use a cached version.
+
+        Finally, do the actual conversion and export the converted file.
+        """
```
birdbrained: Focus the logic in the handler so its clearer what this endpoint does.

mfr/server/handlers/export.py
```diff
@@ -37,61 +37,49 @@ class ExportHandler(core.BaseHandler):
...
         else:
             cache_file_path_str = '/export/{}'.format(self.cache_file_id)
-        self.cache_file_path = await self.cache_provider.validate_path(cache_file_path_str)
-
-        self.source_file_path = await self.local_cache_provider.validate_path(
-            '/export/{}'.format(self.source_file_id)
-        )
-
-        self.output_file_id = '{}.{}'.format(self.source_file_path.name, self.format)
-        self.output_file_path = await self.local_cache_provider.validate_path(
```
birdbrained: need to make sure all unnecessary path stiff is removed. Only remote cache and osf provider should remain.

mfr/server/handlers/export.py
```diff
@@ -37,61 +37,49 @@ class ExportHandler(core.BaseHandler):
...
+
+        # Try to get a cached version
         if settings.CACHE_ENABLED:
             try:
-                cached_stream = await self.cache_provider.download(self.cache_file_path)
+                self.cache_file_path = await self.cache_provider.validate_path(cache_file_path_str)
+                await self.write_stream(self.cache_provider.download(self.cache_file_path))
+                logger.info('Cached file found; Sending downstream [{}]'.format(self.cache_file_path))
+                self.metrics.add('cache_file.result', 'hit')
+                return
```
birdbrained: clean up the cache try.

test.py
```diff
@@ -0,0 +1,14 @@
+def raiser():
+    raise "ERR"
+
+def test():
+    try:
```
birdbrained: oops need to remove.

mfr/server/handlers/render.py
```diff
@@ -32,59 +31,47 @@ class RenderHandler(core.BaseHandler):
...
-        loop = asyncio.get_event_loop()
-        rendition = await loop.run_in_executor(None, renderer.render)
-        self.renderer_metrics = renderer.renderer_metrics
+        # Perform the render and write the result to the response
+        await self.write_stream(await render())
 
-        # Spin off upload into non-blocking operation
-        if renderer.cache_result and settings.CACHE_ENABLED:
+        # Spin off upload of cached render into non-blocking operation
+        if render.cache_result and settings.CACHE_ENABLED:
```
birdbrained: this doesn't line up with what the exporter does to cache. Should decide on a similar approach and use in both handlers.

mfr/core/extension.py
```diff
@@ -1,11 +1,25 @@
...
+class Cacheable(metaclass=abc.ABCMeta):
 
+    @abc.abstractproperty
+    def cache_result(self):
+        pass
 
 class BaseExporter(metaclass=abc.ABCMeta):
 
-    def __init__(self, ext, source_file_path, output_file_path, format):
+    def __init__(self, metadata, input_stream, format):
```
birdbrained: pass the stream in to the exporter so the subcalsses' export method can choose to use the stream rather than writing to disk if it is feasable and appropriate.

birdbrained:
This probably breaks a bunch of tests. I want to make sure we like the implementation before spending time fixing tests.

mfr/server/handlers/render.py
cslzchen: Why not using the name **renderer**? I mean `renderer = utils.bind_render()`?

cslzchen: Before the flow is: render -> future cache -> write. Now it is render -> write -> future cache? But the `rendition` is not longer available here.

mfr/core/extension.py
cslzchen: Similar to renderer, are there any side effects for let renderer/exporter handle `asyncio`? Is it possible to move this back to the handler but still keep your code move. I feel more comfortable for the handlers to take care of the asynchronous business all in one place stead of delegating it to the exporter instance.

mfr/core/extension.py
cslzchen: I think I mentioned this somewhere above, is this really necessary? Why context manager doesn't work in your case?

mfr/core/extension.py
```diff
@@ -15,38 +29,90 @@ def __init__(self, ext, source_file_path, output_file_path, format):
...
+
+    def __del__(self):
+        self.output_fp.close()
+
     @abc.abstractmethod
     def export(self):
         pass
 
+    async def write_to_stream(self):
+        self.output_fp = open(self.output_file_path, 'rb')
```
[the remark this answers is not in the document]
birdbrained: this needs to keep the fp alive even after the fn returns. the context manager would need to be inside the handler, and this class will need to define `__enter__` and `__leave__`
```
with render() as rendered:
    self.write_stream(rendered)
```

mfr/core/utils.py
```diff
@@ -63,6 +63,51 @@ def make_exporter(name, source_file_path, output_file_path, format):
...
+            invoke_args={
+                'metadata': metadata.serialize(),
+                'file_stream': file_stream,
+                'url': url,
+                'assets_url': assets_url,
+                'export_url': export_url,
+            }
+        )
+
+def bind_convert(metadata, file_stream, format):
```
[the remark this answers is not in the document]
birdbrained: The Exporter constructor is basically binding the arguments to a function. Imagine a fun called bind that behaves like this:
```
>>> def hello(msg):
...     print(msg)
...
>>> bound_fn = bind(hello, "Hello")
>>> bound_fn()
Hello

```

mfr/server/handlers/export.py
```diff
@@ -37,61 +37,49 @@ class ExportHandler(core.BaseHandler):
             cache_file_path_str = '/export/{}.{}'.format(self.cache_file_id, self.exporter_name)
         else:
             cache_file_path_str = '/export/{}'.format(self.cache_file_id)
-        self.cache_file_path = await self.cache_provider.validate_path(cache_file_path_str)
```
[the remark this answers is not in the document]
birdbrained: this should not be in the exporter - the exporter doesn't do caching it should only be in the handler.

mfr/server/handlers/export.py
```diff
@@ -37,61 +37,49 @@ class ExportHandler(core.BaseHandler):
...
-            'output_file': {
-                'id': self.output_file_id,
-                'path': str(self.output_file_path),
-                'provider': self.local_cache_provider.NAME,
-            }
-        })
 
     async def get(self):
-        """Export a file to the format specified via the associated extension library"""
+        """Export a file to the format specified via the associated extension
```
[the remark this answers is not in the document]
birdbrained: hm this seems odd.

mfr/server/handlers/export.py
```diff
@@ -37,61 +37,49 @@ class ExportHandler(core.BaseHandler):
...
                 self.metrics.add('cache_file.result', 'miss')
-            else:
-                logger.info('Cached file found; Sending downstream [{}]'.format(self.cache_file_path))
-                self.metrics.add('cache_file.result', 'hit')
-                self._set_headers()
-                return await self.write_stream(cached_stream)
 
-        await self.local_cache_provider.upload(
+        # File isn't cached and it needs to be converted
+        await self.write_stream(await utils.bind_convert(
```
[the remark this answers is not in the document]
birdbrained: While this is technically true, adding support for the noop also required some changes to `prepare`  because some objects were no longer necessary. Making that logic work inside of the exporter rather than the handler turned out to be a little simpler to write.

mfr/server/handlers/render.py
```diff
@@ -32,59 +31,47 @@ class RenderHandler(core.BaseHandler):
...
             self.request.uri.replace('/render?', '/export?', 1)
         )
 
-        self.extension_metrics.add('class', renderer._get_module_name())
-
-        if renderer.cache_result and settings.CACHE_ENABLED:
+        # Try and use a cached version of the render
+        if render.cache_result and settings.CACHE_ENABLED:
             try:
+                self.cache_file_path = await self.cache_provider.validate_path(cache_file_path_str)
```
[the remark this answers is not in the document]
birdbrained: Will take a look at this. I didn't know it should be a goal to minimize the code inside of a `try`

mfr/server/handlers/render.py
birdbrained: render - the action of rendering, renderer - a thing that renders, render is a callable that performs the action of rendering. Can't use `renderer.render`

birdbrained: I'm not sure this order matters. I placed the ensure_future after to help visualize that this is something that temporally happens after writing the response. I do need a tweak here to store the stream in a local variable for the response and the cache.

mfr/server/handlers/export.py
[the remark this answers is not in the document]
birdbrained: It's become neccesary because we require the id inside the exporter  now.

mfr/core/extension.py
```diff
@@ -72,6 +144,48 @@ def __init__(self, metadata, file_path, url, assets_url, export_url):
         except AttributeError:
             pass
 
+    @property
+    def local_cache_provider(self):
+        if not self._local_cache_provider:
+            self._local_cache_provider = waterbutler.core.utils.make_provider(
+                'filesystem', {}, {}, settings.LOCAL_CACHE_PROVIDER_SETTINGS
+            )
+        return self._local_cache_provider
```
birdbrained: needs to be try/except

mfr/core/extension.py
```diff
@@ -72,6 +144,48 @@ def __init__(self, metadata, file_path, url, assets_url, export_url):
...
+
+    async def get_source_file_path(self):
+        if self._source_file_path is None:
+            self._source_file_path = await self.local_cache_provider.validate_path(
+                '/render/{}'.format(self.source_file_id)
+            )
+        return self._source_file_path
```
birdbrained: needs to be try/except

mfr/core/extension.py
```diff
@@ -72,6 +144,48 @@ def __init__(self, metadata, file_path, url, assets_url, export_url):
...
+
+    async def __call__(self):
+
+        self.renderer_metrics.add('class', self._get_module_name())
+
+        if self.file_required:
+            await self.local_cache_provider.upload(
+                await self.file_stream,
+                await self.source_file_path
+            )
```
birdbrained: should be irrelevant

mfr/core/extension.py
```diff
@@ -72,6 +144,48 @@ def __init__(self, metadata, file_path, url, assets_url, export_url):
...
+        rendition = self.render()
+
+        return StringStream(rendition)
+
+    def __del(self):
+        if self.file_required:
+            try:
+                remove(self._source_file_path.full_path)
+            except FileNotFoundError:
+                pass
```
birdbrained: Should be irrelevant

mfr/core/utils.py
```diff
@@ -33,73 +33,52 @@ def make_provider(name, request, url):
...
-
-    :param str name: The name of the extension to instantiate. (.jpg, .docx, etc)
-    :param str source_file_path:
-    :param str output_file_path:
-    :param str format:
-
-    :rtype: :class:`mfr.core.extension.BaseExporter`
-    """
-    normalized_name = (name and name.lower()) or 'none'
+def bind_render(metadata, file_stream, url, assets_url, export_url):
```
birdbrained: Needs docstring

mfr/core/utils.py
```diff
@@ -33,73 +33,52 @@ def make_provider(name, request, url):
...
-    :param: :class:`mfr.core.provider.ProviderMetadata` metadata:
-    :param str file_path:
-    :param str url:
-    :param str assets_url:
-    :param str export_url:
-
-    :rtype: :class:`mfr.core.extension.BaseRenderer`
-    """
-    normalized_name = (name and name.lower()) or 'none'
+def bind_convert(metadata, file_stream, format):
```
birdbrained: can theses two utils become one?

mfr/extensions/image/export.py
```diff
@@ -16,53 +20,62 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.metrics.add('pil_version', Image.VERSION)
 
-    def export(self):
+    async def export(self):
```
birdbrained: this exporter feels a little messy still

mfr/extensions/jsc3d/export.py
```diff
@@ -1,44 +1,64 @@
...
-        try:
-            # alter original source file with a named temp file, so freecad can read it properly
-            with NamedTemporaryFile(mode='w+b') as temp_source_file:
-                temp_source_file.name = self.source_file_path + '.step'
-                shutil.copy2(self.source_file_path, temp_source_file.name)
+    async def export(self):
+        # alter original source file with a named temp file, so freecad can read it properly
+        with NamedTemporaryFile(mode='w+b') as temp_source_file:
+            temp_source_file.name = await self.source_file_path + '.step'
+            shutil.copy2(await self.source_file_path, temp_source_file.name)
```
birdbrained: Instead of copy, can we hardlink? assuming name of file is the issue

mfr/server/handlers/core.py
```diff
@@ -119,21 +130,35 @@ def NAME(self):
...
-
-        self.local_cache_provider = waterbutler.core.utils.make_provider(
-            'filesystem', {}, {}, settings.LOCAL_CACHE_PROVIDER_SETTINGS
-        )
-
         self.source_file_id = uuid.uuid4()
         self.add_header('X-MFR-REQUEST-ID', str(uuid.uuid4()))
 
+    @property
+    def plugin_name(self):
```
birdbrained: Needs a docstring

mfr/server/handlers/export.py
```diff
@@ -65,58 +56,45 @@ class ExportHandler(core.BaseHandler):
             self.metrics.add('export.conversion', 'noop')
             return
 
+        # Try to get a cached version
         if settings.CACHE_ENABLED:
             try:
-                cached_stream = await self.cache_provider.download(self.cache_file_path)
+                await self.write_stream(self.cache_provider.download(self.cache_file_path))
+                logger.info('Cached file found; Sending downstream [{}]'.format(self.cache_file_path))
+                self.metrics.add('cache_file.result', 'hit')
```
birdbrained: Bread into handler method

mfr/server/handlers/export.py
```diff
@@ -65,58 +56,45 @@ class ExportHandler(core.BaseHandler):
...
+                return
+
+            # Cache miss
```
birdbrained: Break into handler method

mfr/server/handlers/render.py
```diff
@@ -12,83 +12,149 @@
...
 logger = logging.getLogger(__name__)
 
+DEFAULT_RENDER = {'renderer': '.pdf', 'format': 'pdf'}
+
+RENDER_MAP = {
+    # 'csv': {'renderer': '.xlsx', 'format': 'xlsx'},
+    # 'ppt': {'renderer': '.pdf', 'format': 'pdf'},
+    # 'pptx': {'renderer': '.pdf', 'format': 'pdf'},
+}
+
```
birdbrained: Move this

mfr/server/handlers/render.py
```diff
@@ -12,83 +12,149 @@
...
 
 class RenderHandler(core.BaseHandler):
```
birdbrained: Needs a docstring

mfr/server/handlers/render.py
```diff
@@ -12,83 +12,149 @@
...
+        needed.
+        """
+        try:
+            return self._source_stream
+        except:
+            self._source_stream = self.provider.download()
+            return self._source_stream
+
+    @property
+    def export_url(self):
```
birdbrained: Needs a docstring

mfr/server/handlers/render.py
```diff
@@ -12,83 +12,149 @@
...
+        from the original waterbutler-sourced file to the future returned by
+        exporting the file.
+
+        Also change the url for the file to an export url
+        for the case that the file is not required."""
+        try:
+            map = RENDER_MAP[self.metadata.ext]
+        except KeyError:
+            map = DEFAULT_RENDER
+        self._source_stream = utils.bind_convert(
```
birdbrained: make this a lambda(?) and handle the case in the property so it's not called unless needed

mfr/server/handlers/render.py
```diff
@@ -12,83 +12,149 @@
...
 
-        await self.write_stream(waterbutler.core.streams.StringStream(rendition))
-
-    async def _cache_and_clean(self):
-        if hasattr(self, 'source_file_path'):
-            try:
-                os.remove(self.source_file_path.full_path)
-            except FileNotFoundError:
-                pass
+    def cache_result(self, rendition):
```
birdbrained: Needs a docstring

setup.py
```diff
@@ -724,78 +724,78 @@ def parse_requirements(requirements):
...
+            '.dif = mfr.extensions.pdf:PdfRenderer',
+            '.doc = mfr.extensions.pdf:PdfRenderer',
+            '.docx = mfr.extensions.pdf:PdfRenderer',
+            '.emf = mfr.extensions.pdf:PdfRenderer',
+            '.eps = mfr.extensions.pdf:PdfRenderer',
+            '.fodg = mfr.extensions.pdf:PdfRenderer',
+            '.fodp = mfr.extensions.pdf:PdfRenderer',
+            '.fods = mfr.extensions.pdf:PdfRenderer',
+            '.fodt = mfr.extensions.pdf:PdfRenderer',
+            # '.gif = mfr.extensions.pdf:UnoconvRenderer',
```
birdbrained: Fix all of them

PR #339, opened 2018-05-15
[SVCS-612] Add Tabular error for logs

birdbrained:
## Ticket
https://openscience.atlassian.net/browse/SVCS-612

## Purpose

This changes the logging message to include the name of the error that actually triggered it to assist in debugging tasks

## Changes

Adds the nae of the error to the log message

## Side effects

Better debugging

### DIFF
--- a/mfr/extensions/tabular/render.py
+++ b/mfr/extensions/tabular/render.py
@@ -119,7 +119,9 @@ class TabularRenderer(extension.BaseRenderer):
                 self._renderer_tabular_metrics['importer'] = function.__name__
                 try:
                     return imported(fp)
-                except (KeyError, ValueError):
+                except (KeyError, ValueError) as err:
+                    logger.error('WB has encountered an unexpected error '
+                                 'when trying to render a tabular file: {}'.format(err))
                     raise exceptions.UnexpectedFormattingError(
                         'Unexpected formatting error.',
                         extension=self.metadata.ext,
--- a/mfr/extensions/tabular/render.py
+++ b/mfr/extensions/tabular/render.py
@@ -110,13 +110,13 @@ class TabularRenderer(extension.BaseRenderer):
         """
         function_preference = settings.LIBS.get(ext.lower())
 
-        for function in function_preference:
+        for populate_func in function_preference:
             try:
-                imported = function()
+                imported = populate_func()
             except ImportError:
                 pass
             else:
-                self._renderer_tabular_metrics['importer'] = function.__name__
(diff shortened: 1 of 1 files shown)


mfr/extensions/tabular/render.py
```diff
@@ -119,9 +119,9 @@ def _populate_data(self, fp, ext):
                 self._renderer_tabular_metrics['importer'] = function.__name__
                 try:
                     return imported(fp)
-                except (KeyError, ValueError):
+                except (KeyError, ValueError) as err:
                     raise exceptions.UnexpectedFormattingError(
-                        'Unexpected formatting error.',
```
birdbrained: Add the message too

mfr/extensions/tabular/render.py
```diff
@@ -119,9 +119,9 @@ def _populate_data(self, fp, ext):
...
+                        'Unexpected formatting error. {!r}'.format(err),
```
cslzchen: 👍 for `{!r}`

cslzchen:
A better example

### With `{!r}`

screen shot 2018-06-12 at 2 25 45 pm

```bash
mfr.extensions.tabular.exceptions.UnexpectedFormattingError: 500, Unexpected formatting error. <EmptyTableError(400, Table empty or corrupt.)>
```

### With {!s}

screen shot 2018-06-12 at 2 27 02 pm

```bash
mfr.extensions.tabular.exceptions.UnexpectedFormattingError: 500, Unexpected formatting error. 400, Table empty or corrupt.
```

cslzchen:
As discussed with @birdbrained , we don't want the detailed error message goes to the user. So (1) the general error message remains the same (2) send a log to sentry (`sentry.log()`???) with detailed error message and stack trace.

mfr/extensions/tabular/render.py
```diff
@@ -119,7 +119,8 @@ def _populate_data(self, fp, ext):
                 self._renderer_tabular_metrics['importer'] = function.__name__
                 try:
                     return imported(fp)
-                except (KeyError, ValueError):
+                except (KeyError, ValueError) as err:
+                    logger.error('Encountered an unexpected error when trying to render a table: {}'.format(err))
```
cslzchen: Fyi, my memory has failed me. I remember that `logger.error()` doesn't go to Sentry and that we need to use `sentry.log_message()`, which turns out to be my old memories about OSF. I might have givne you some half-wrong advice. 👍 for figure this out.

PR #340, opened 2018-05-15
[SVCS-592] Fix idiotic bug

birdbrained:
Make the way image dimensions are pulled out of paths more robust

## Ticket

https://openscience.atlassian.net/browse/SVCS-592

## Purpose

Make an idiotic bug not happen. Changes the way that image sizes are pulled out of the url.

### DIFF
--- a/mfr/extensions/image/export.py
+++ b/mfr/extensions/image/export.py
@@ -19,11 +19,16 @@ class ImageExporter(extension.BaseExporter):
     def export(self):
         parts = self.format.split('.')
         type = parts[-1].lower()
-        max_size = [int(x) for x in parts[0].split('x')] if len(parts) == 2 else None
+        max_size = {
+            'w': None,
+            'h': None
+        }
+        if len(parts) == 2:
+            max_size['w'], max_size['h'] = [int(size) for size in parts[0].split('x')]
         self.metrics.merge({
             'type': type,
-            'max_size_w': max_size[0],
-            'max_size_h': max_size[1],
+            'max_size_w': max_size['w'],
+            'max_size_h': max_size['h'],
         })
         try:
             if self.ext in ['.psd']:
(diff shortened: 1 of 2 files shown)


mfr/extensions/image/export.py
```diff
@@ -38,7 +43,7 @@ def export(self):
 
             if max_size:
```
birdbrained: Check w and h explicitly

mfr/extensions/image/export.py
```diff
@@ -19,11 +19,16 @@ def __init__(self, *args, **kwargs):
     def export(self):
         parts = self.format.split('.')
         type = parts[-1].lower()
-        max_size = [int(x) for x in parts[0].split('x')] if len(parts) == 2 else None
+        max_size = {
+            'w': None,
```
cslzchen: Not a blocker but how about using `width` and `height` instead of `w` and `h`?

mfr/extensions/image/export.py
```diff
@@ -36,13 +41,18 @@ def export(self):
             else:
                 image = Image.open(self.source_file_path)
 
-            if max_size:
+            if any(size is None for _, size in max_size.items()):
```
cslzchen: `max_size` is dictionary containing only two keys `w` and `h`. This `if` makes things look complicated. Why not use:

```python
if max_size.get('w') and max_size.get('h'):
    ....
```

In addition, if there is a real issue that one of the two is missing (, which I doubt), we should raise an error as well.

setup.cfg
```diff
@@ -9,3 +9,6 @@
 ignore = E501,E127,E128,E265,E301,E302,F403,E731
 max-line-length = 100
 exclude = .ropeproject,tests/*,src/*,env,venv,node_modules/*
+
+[pytest]
+norecursedirs = .* build CVS _darcs {arch} *.egg venv node_modules/*
```
cslzchen: Unrelated changes. Please remove or make another PR for it.

cslzchen:
I merged #339 to local `develop`, rebased #340 on the local `develop` and pushed back to the PR in order to have the same commit without pushing local `develop` up twice. A side effect is the PR commit history looks bad. Next time I will merge them without rebasing on each other. cc @felliott

PR #344, opened 2018-06-20
Ignore node_modules when running pytest

birdbrained:
There are some python files in node_modeles, and these packages tests
shouldn't run when we run MFR tests.


birdbrained:
This is the same as default, only it adds the config not to descend into node_modules directories when looking for tests to run. Some node modules have python tests. This may only apply if you run py.test directly, but I do that, so.

PR #349, opened 2018-08-06
Dont pass empty header

birdbrained:
## Ticket

No ticket

## Purpose

Tidies up the operation of a little bit of code.

## Changes

The OSF provider was passing a header with an empty value, this change
means the if the header is empty, then the header isn't sent at all.

This technically is a small optimization as well, it removes a few
instructions that would get called inside aiohttp. if this header was
here. (This is probably insignificant, however.)

## Side effects

OSF gets no header in this situation now

### DIFF
--- a/mfr/providers/osf/provider.py
+++ b/mfr/providers/osf/provider.py
@@ -73,7 +73,7 @@ class OsfProvider(provider.BaseProvider):
             metadata_response = await self._make_request(
                 'HEAD',
                 download_url,
-                headers={settings.MFR_ACTION_HEADER: self.action or ''}
+                headers=({settings.MFR_ACTION_HEADER: self.action} if self.action else None)
             )
             response_code = metadata_response.status
             response_reason = metadata_response.reason


mfr/providers/osf/provider.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, request, url, action=None):
             metadata_response = await self._make_request(
                 'HEAD',
                 download_url,
-                headers={settings.MFR_ACTION_HEADER: self.action or ''}
+                headers=({settings.MFR_ACTION_HEADER: self.action} if self.action else None)
```
cslzchen: * I don't think you need the extra `()`. This should be good.

```python
headers={settings.MFR_ACTION_HEADER: self.action} if self.action else None
```

* A header with empty value is different from not having the header. [Here](https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/auth/osf/handler.py#L108) is how WB uses this header. It shouldn't be sent if it is empty. 👍 

```python
...
mfr_action = request.headers[settings.MFR_ACTION_HEADER].lower()
try:
    osf_action = mfr_action_map[mfr_action]
...
```

PR #353, opened 2018-09-27
Remove rackspace utilities

birdbrained:
https://openscience.atlassian.net/browse/SVCS-896

## Purpose
Remove rackspace cloudfiles tasks from MFR's `tasks.py` and remove the pyrax dependency. We no longer use RSCF for the MFR cache. Rackspace is no longer used, so pyrax is an unneccesary dep. and the tasks referring to it can be removed.


birdbrained:
Oh I think you're right. I'll update this pr to put those in.

birdbrained:
I cannot find anything else related in MFR. If anyone knows of anything else that needs to be in this PR, fell free to note where that is and move it back to addnl. dev. For now, I'll move it back to CR.

PR #355, opened 2019-03-21
[ENG-44] Add google analytics tracking for when hypothesis side panel is opened

adlius:
## Ticket
https://openscience.atlassian.net/browse/ENG-44

## Purpose

To add GA tracking to MFR to track when users click open the Hypothesis panel.

## Changes
- In `mfr/extensions/pdf/templates/viewer.mako`, we add GA tracking imports.
- In `mfr/server/static/js/mfr.child.hypothesis.js`, we set `window.hypothesisConfig` to use the `onLayoutChange()` callback provided by hypothesis to watch for opening of annotation panel.
- Add GA tracking ID to settings and use that in renderer.

## Side effects

None

## QA Notes

Check to make sure the event is only tracked when users open the panel for the first time on that page. Page refresh counts as a new page. User is counted by browser, not OSF User. So if the same OSFUser opens two browser and opens the panel in both, it counts as two events.

## Deployment Notes

Add GOOGLE_ANALYTICS_TRACKING_ID for to staging/prod settings.

### DIFF
--- a/mfr/extensions/pdf/render.py
+++ b/mfr/extensions/pdf/render.py
@@ -7,6 +7,7 @@ from mako.lookup import TemplateLookup
 from mfr.core import extension
 from mfr.extensions.pdf import settings
 from mfr.extensions.utils import munge_url_for_localdev, escape_url_for_template
+from mfr.settings import GOOGLE_ANALYTICS_TRACKING_ID
 
 logger = logging.getLogger(__name__)
 
@@ -29,6 +30,7 @@ class PdfRenderer(extension.BaseRenderer):
         if self.metadata.ext.lower() not in settings.EXPORT_SUPPORTED:
             logger.debug('Extension not found in supported list!')
             return self.TEMPLATE.render(
+                ga_tracking_id=GOOGLE_ANALYTICS_TRACKING_ID,
                 base=self.assets_url,
                 url=escape_url_for_template(download_url.geturl()),
                 stable_id=self.metadata.stable_id,
@@ -46,6 +48,7 @@ class PdfRenderer(extension.BaseRenderer):
 
         self.metrics.add('needs_export', True)
         return self.TEMPLATE.render(
+            ga_tracking_id=GOOGLE_ANALYTICS_TRACKING_ID,
             base=self.assets_url,
             url=escape_url_for_template(exported_url.url),
             stable_id=self.metadata.stable_id,
--- a/mfr/server/static/js/mfr.child.hypothesis.js
+++ b/mfr/server/static/js/mfr.child.hypothesis.js
@@ -64,13 +64,24 @@
             hypothesisLoaded = true;
 
             var sidePanelOpened = false;
+            // window.DEFAULT_URL should be the wb link in this format:
+            // https://<wb-domain>/v1/resources/<preprint-guid>/providers/osfstorage/<file-id>?direct=&mode=render
+            // TODO: parse and validate the WB URL before retrieving the preprints GUID
+            var wbLink = window.DEFAULT_URL;
+            var preprintGuid;
+            if (wbLink.split('/').length >= 6) {
+                preprintGuid = wbLink.split('/')[5];
+            } else {
+                preprintGuid = 'preprint-guid-unknown';
+            }
             var sendAnalyticsIfExpanded = function (expanded) { 
                 if (expanded && !sidePanelOpened) {
                     sidePanelOpened = expanded;
                     ga('send', 'event', {
                         eventCategory: 'Hypothesis',
                         eventAction: 'Open Hypothesis Panel',
-                        eventLabel: window.DEFAULT_URL.split('/')[5],
+                        //`eventLabel` is the guid of the preprint to which the file belongs
+                        eventLabel: preprintGuid,
                     });
                 }
             };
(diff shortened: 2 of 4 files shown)


mfr/server/static/js/mfr.child.hypothesis.js
```diff
@@ -62,7 +62,23 @@
             window.document.head.appendChild(script);
             window.document.body.classList.add('show-hypothesis');
             hypothesisLoaded = true;
-        });
 
+            var sidePanelOpened = false;
+            var sendAnalyticsIfExpanded = function (expanded) { 
+                if (expanded && !sidePanelOpened) {
```
cslzchen: Probably a dumb question about `hypothesis` configuration. What is `expanded` (or `layout.expanded` a few lines  below more specifically)?

mfr/settings.py
```diff
@@ -170,3 +170,4 @@ def child(key):
 
 
 SENTRY_DSN = config.get_nullable('SENTRY_DSN', None)
+GOOGLE_ANALYTICS_TRACKING_ID = config.get_nullable('GOOGLE_ANALYTICS_TRACKING_ID', None)
```
cslzchen: Can GA script handle `None` tracking ID? I guess so but just want to double check with you.

mfr/extensions/pdf/render.py
```diff
@@ -5,6 +5,7 @@
 from mako.lookup import TemplateLookup
 
 from mfr.core import extension
+from mfr.settings import GOOGLE_ANALYTICS_TRACKING_ID
```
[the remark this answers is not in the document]
aaxelb: Drive-by: That sound worth enforcing
https://github.com/PyCQA/flake8-import-order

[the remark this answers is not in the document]
cslzchen: Such a nice feature I have always wanted from `flake8`. 👍

mfr/settings.py
adlius: I am not sure. It'll probably throw an error.

mfr/server/static/js/mfr.child.hypothesis.js
adlius: `layout.expanded` denotes whether the hypothesis side panel is expanded when `onLayoutChanged()` is called.

mfr/settings.py
adlius: @cslzchen Confirmed that it would not cause an error when tracking ID is `None`.

mfr/server/static/js/mfr.child.hypothesis.js
```diff
@@ -62,7 +62,23 @@
...
+                    sidePanelOpened = expanded;
+                    ga('send', 'event', {
+                        eventCategory: 'Hypothesis',
+                        eventAction: 'Open Hypothesis Panel',
+                        eventLabel: window.DEFAULT_URL.split('/')[5],
```
felliott: What is this value?  Could you add a comment saying what this is expected to be?

adlius: `eventLabel` should be the preprint guid.

adlius:
@felliott Comment is added.

mfr/server/static/js/mfr.child.hypothesis.js
cslzchen: A follow-up question, is it guaranteed that the `GUID` is always at index 5? Can we add some extra check before `ga` to parse `window.DEFAULT_URL.split('/')[5]` and handle unexpected URL format. In addition, having a comment on what the expected URL looks like will be very helpful as well.

adlius: There is no guarantee that `window.DEFAULT_URL.split('/')[5]` always return the preprint guid. However, if the rendered file is a preprint's primary file, `window.DEFAULT_URL` should be the WB link to the file on OSFStorage (e.g. https://files.us.staging2.osf.io/v1/resources/twq6d/providers/osfstorage/5c9cd1e3c8f5970017ed07d3?direct=&mode=render), and from which we can extract the guid of the preprint.
And since hypothesis is only enabled on a preprint's primary file, it should be fine.
I'll add some checks for unexpected URL and log to sentry when that happens.

cslzchen: Sounds good! Thanks for the explanation. This WB API rarely changes. If it ever changed, we would be informed right away.

adlius: @cslzchen PR updated.

cslzchen: I added a TODO comment to your code. Although the `split` you use here is unaware of the HTTP protocol, URL domain and path, I think for now it is good enough. A better way (probably a future improvement) is to fully parse and validate the URL.

mfr/server/static/js/mfr.child.hypothesis.js
```diff
@@ -62,7 +62,34 @@
             window.document.head.appendChild(script);
             window.document.body.classList.add('show-hypothesis');
             hypothesisLoaded = true;
-        });
 
+            var sidePanelOpened = false;
+            // window.DEFAULT_URL should be the wb link in this format:
+            // https://<wb-domain>/v1/resources/<preprint-guid>/providers/osfstorage/<file-id>?direct=&mode=render
+            // TODO: parse and validate the WB URL before retrieving the preprints GUID
```
cslzchen: I added a TODO comment to your code. Although the `split` you use here is unaware of the HTTP protocol, URL domain and path, I think for now it is good enough. A better way (probably a future improvement) is to fully parse and validate the URL.

PR #358, opened 2019-09-17
[CR][ENG-379] update Dockerfile to be based on debian buster

felliott:
## Ticket

https://openscience.atlassian.net/browse/ENG-379

## Purpose

Update base image to Debian Buster to allow installing LibreOffice from apt-get.

## Changes

Install libreoffice via apt-get instead of downloading `.deb` file from LO's archives.

## Side effects

None expected.  This can be tested/released independently of the [`docker-library` fix](https://github.com/CenterForOpenScience/docker-library/pull/51).

## QA Notes

Dev QA only.

## Deployment Notes

Nope.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,12 @@
-FROM python:3.5-slim-stretch
+FROM python:3.5-slim-buster
 
 # ensure unoconv can locate the uno library
-ENV PYTHONPATH=/usr/lib/python3/dist-packages
+ENV PYTHONPATH /usr/lib/python3/dist-packages
 
 RUN usermod -d /home www-data \
     && chown www-data:www-data /home \
+    # -slim images strip man dirs, but java won't install unless this dir exists.
+    && mkdir -p /usr/share/man/man1 \
     && apt-get update \
     # mfr dependencies
     && apt-get install -y \
@@ -27,6 +29,8 @@ RUN usermod -d /home www-data \
         freecad \
         # pspp dependencies
         pspp \
+        # unoconv dependencies
+        libreoffice \
         # gosu dependencies
         curl \
         gnupg2 \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,10 +1,12 @@
-FROM python:3.5-slim-stretch
+FROM python:3.5-slim-buster
 
 # ensure unoconv can locate the uno library
-ENV PYTHONPATH=/usr/lib/python3/dist-packages
+ENV PYTHONPATH /usr/lib/python3/dist-packages
```
cslzchen: I was wondering whether this was a bug but learned that bot (w/ or w/o `=`) are valid Dockerfile statement for setting an `ENV`: https://docs.docker.com/engine/reference/builder/#env.

Dockerfile
```diff
@@ -1,10 +1,12 @@
...
 
 RUN usermod -d /home www-data \
     && chown www-data:www-data /home \
+    # -slim images strip man dirs, but java won't install unless this dir exists.
+    && mkdir -p /usr/share/man/man1 \
```
cslzchen: Yep, I searched a little bit and discovered that this was and still is bug for jre on slim debian. And I learned the `-p` option for `mkdir`:

> -p, --parents: no error if existing, make parent directories as needed

Dockerfile
```diff
@@ -53,37 +57,8 @@ RUN usermod -d /home www-data \
     && rm -rf /var/lib/apt/lists/* \
     && pip install -U pip \
     && pip install setuptools==37.0.0 \
-    && mkdir -p /code
-
-ENV LIBREOFFICE_VERSION 6.1.6
```
cslzchen: One note: the diff is misleading. I think the change here is removing installing Libreoffice from Libreoffice here but from Debian using `apt-get` many lines above.

PR #362, opened 2020-05-29
[WIP] [ForestFire] Improve .csv dialect sniffing

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/SVCS-531

## Purpose

Improve .csv dialect sniffing.

## Changes

* Provide a full row of data to `csv.Sniffer().sniff()` so it can effectively detect the correct dialect. MFR reads a small amount of data from the file and recursively read more until either the following happens.
  * The sniff size goes over max render size allowed -> return renderer error
  * Any type of new line character found, trim (if necessary) and send the the full first row to the sniffer.

* - [ ] TODO: update unit tests

## Side effects

Memory usage for sniffing depends on the size of the first row of the file. In the worst case, it could sniff at most 10MB. Please note that we don't have enough statistical data on how many bytes the first row of a `.csv` file takes on average.

- [ ] TODO: maybe add a sentry alert when the sniffing size goes over 512KB

## QA Notes

TBD

## Deployment Notes

N / A

### DIFF
--- a/mfr/extensions/tabular/libs/stdlib_tools.py
+++ b/mfr/extensions/tabular/libs/stdlib_tools.py
@@ -1,18 +1,32 @@
 import re
 import csv
+import logging
 
-from mfr.extensions.tabular.exceptions import EmptyTableError, TabularRendererError
 from mfr.extensions.tabular import utilities
+from mfr.extensions.tabular.settings import MAX_FILE_SIZE, TABULAR_INIT_SNIFF_SIZE
+from mfr.extensions.tabular.exceptions import EmptyTableError, TabularRendererError
+
+logger = logging.getLogger(__name__)
 
 
 def csv_stdlib(fp):
     """Read and convert a csv file to JSON format using the python standard library
-    :param fp: File pointer object
-    :return: tuple of table headers and data
+
+    Quirk: ``csv.Sniffer().sniff()`` needs the FULL first row and ONLY one full row to be able to
+    effectively detect the correct dialect of the file.
+
+    :param fp: the file pointer object
+    :return: a tuple of table headers and data
     """
-    data = fp.read(2048)
+
+    # Prepare the first row for sniffing
+    data = fp.read(TABULAR_INIT_SNIFF_SIZE)
+    data = _trim_or_append_data(fp, data, TABULAR_INIT_SNIFF_SIZE, 0)
+
+    # Reset the file pointer
     fp.seek(0)
 
+    # Sniff the first row to find a matching format
     try:
         dialect = csv.Sniffer().sniff(data)
     except csv.Error:
@@ -20,10 +34,14 @@ def csv_stdlib(fp):
     else:
         _set_dialect_quote_attrs(dialect, data)
 
+    # Explicitly delete data when it is on longer used.
     del data
+
+    # Create the CSV reader with the detected dialect
     reader = csv.DictReader(fp, dialect=dialect)
+
+    # Update the reader field names to avoid duplicate column names when performing row extraction
     columns = []
-    # update the reader field names to avoid duplicate column names when performing row extraction
     for idx, fieldname in enumerate(reader.fieldnames or []):
         column_count = sum(1 for column in columns if fieldname == column['name'])
         if column_count:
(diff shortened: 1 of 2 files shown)


mfr/extensions/tabular/libs/stdlib_tools.py
```diff
@@ -1,29 +1,52 @@
...
+
+    logger.info('>>> ??? &&& ~~~ current settings')
+    logger.info('>>> ??? &&& ~~~ max rendering file size = {}'.format(MAX_FILE_SIZE))
+    logger.info('>>> ??? &&& ~~~ initial sniffing size = {}'.format(TABULAR_INIT_SNIFF_SIZE))
+
+    # Prepare the first row for sniffing
+    data = fp.read(TABULAR_INIT_SNIFF_SIZE)
+    data = _trim_or_append_data(fp, data, TABULAR_INIT_SNIFF_SIZE, 0)
+
+    # Reset the file pointer
```
cslzchen, line 30: The first solution (see below) didn't work 100% since for one file, sniffing the full file ended up in wrong delimiter while sniffing only the first row worked as expected. This is why even the sniff can sniff the full file, I only provide it with the first row.

```python
...
data = fp.read(TABULAR_INIT_SNIFF_SIZE)
if len(data) == TABULAR_INIT_SNIFF_SIZE:
    data = _trim_or_append_data(fp, data, TABULAR_INIT_SNIFF_SIZE, 0)
...
```

mfr/extensions/tabular/libs/stdlib_tools.py
```diff
@@ -92,3 +115,62 @@ def _set_dialect_quote_attrs(dialect, data):
...
+                    'return the trimmed text'.format(index))
+        return text[:index]
+    # Otherwise, update `sniff_size` and then sniff more (2 times of the last `read_size`) text
+    sniff_size += read_size
+    read_size *= 2
+    more_text = fp.read(read_size)
+
+    # If text to sniff now goes over the max file size limit, raise the renderer error since there
+    # is no need to sniff when the file is already too large to be rendered.
+    if sniff_size + len(more_text) >= MAX_FILE_SIZE:
```
cslzchen, line 149: This is a redundant check since the renderer check file size in advance before handing off to the sniffer. However, I prefer to keep it since it is logically sound and necessary if we look at the recursion by itself.

mfr/extensions/tabular/libs/stdlib_tools.py
```diff
@@ -1,29 +1,47 @@
...
+    effectively detect the correct dialect of the file.
+
+    :param fp: the file pointer object
+    :return: a tuple of table headers and data
     """
-    data = fp.read(2048)
+
+    # Prepare the first row for sniffing
+    data = fp.read(INIT_SNIFF_SIZE)
+    data = _trim_or_append_data(fp, data, INIT_SNIFF_SIZE, 0)
```
cslzchen, line 24: The first solution (see below) didn't work 100% since for one file, sniffing the full file ended up in wrong delimiter while sniffing only the first row worked as expected. This is why even the sniff can sniff the full file, I only provide it with the first row.

```python
...
data = fp.read(TABULAR_INIT_SNIFF_SIZE)
if len(data) == TABULAR_INIT_SNIFF_SIZE:
    data = _trim_or_append_data(fp, data, TABULAR_INIT_SNIFF_SIZE, 0)
...
```

cslzchen, line 24: The initial solution (see below) didn't work 100% since for some file, sniffing the full file ended up in wrong delimiter while sniffing only the first row worked as expected. This is why even when the sniffer can sniff the full file, MFR only provides it with the first row.

```python
...
data = fp.read(TABULAR_INIT_SNIFF_SIZE)
if len(data) == TABULAR_INIT_SNIFF_SIZE:
    data = _trim_or_append_data(fp, data, TABULAR_INIT_SNIFF_SIZE, 0)
...
```

mfr/extensions/tabular/libs/stdlib_tools.py
```diff
@@ -92,3 +110,63 @@ def _set_dialect_quote_attrs(dialect, data):
...
+
+def _trim_or_append_data(fp, text, read_size, size_to_sniff, max_render_size=MAX_FILE_SIZE):
+    """Recursively read data from a file and return its full first row.  The file starts with
+    ``text`` and the file pointer points to the next character immediately after `text`.
+
+    :param fp: the file pointer from which data is read
+    :param text: the current text chunk to check the new line character
+    :param read_size: the last read size when `fp.read()` is called
+    :param size_to_sniff: the accumulated size fo the text to sniff
+    :param max_render_size: the max file size for render
```
cslzchen, line 123: Instead of using `MAX_FILE_SIZE` directly, the `max_render_size=` provides an option for unit tests to use a much smaller size.

mfr/extensions/tabular/libs/stdlib_tools.py
```diff
@@ -92,3 +110,63 @@ def _set_dialect_quote_attrs(dialect, data):
...
+    3. \r       CR
+
+    :param text: the text string to check
+    :return: the index of the first new line character if found. Otherwise, return -1.
+    """
+    index = text.find('\r\n')
+    if index == -1:
+        index = text.find('\n')
+        if index == -1:
+            index = text.find('\r')
```
cslzchen, line 171: I am not sure why `.rfind()` was used in the initial solution. One guess is that it only reads a fixed size for sniff so it wants to include as many role as possible. However, this is no longer the case in my implementation where only the first row is sniffed.

```python
    index = text.rfind('\r\n')
    if index == -1:
        index = text.rfind('\n')
        if index == -1:
            index = text.rfind('\r')
```

PR #369, opened 2022-06-25
jamovi: update to accomodate new manifest location

jonathon-love:
## Purpose

the jamovi file format is placing the manifest in a different location (it turns out putting it at META-INF/MANIFEST.MF makes you the enemy of every email system).

## Changes

trivial changes, it now reads the new path, before falling back to the old path.

## Side effects

NO

## QA Notes

View a .omv file

## Deployment Notes

NA

PR #372, opened 2023-12-06
Handle newer jasp files

RensDofferhoff:
## Purpose

[JASP](https://github.com/jasp-stats/jasp-desktop) files have been updated recently and could no longer be previewed on OSF.

## Changes

This PR removes some superfluous check that allow the file renderer to show both old and new jasp-files. A new test case is added for newer files.


RensDofferhoff:
Any chance this could be merged quickly? 
Anything we need to add/change to comply?

felliott:
Would there be any drawback to using the same approach as the [jamovi PR](#369 )? I'm prepping a release for that, and it seems like they might be similar.  I'd kinda like to keep the extra validation, since a lot of file types are renamed zip files.

I will test it locally with the jamovi fix (adding a check for the new manifest name), but is there anything else that I should be concerned about?  Thank you!

felliott:
The exact fix didn't work since the format of the manifest changed.  I've adapted it a bit to still do some basic validation. This is merged into `develop` and I will cut a release early next week (post-new-years).

felliott:
This has been released!

JorisGoosen:
It works great, merci

PR #374, opened 2025-04-24
[ENG-7597] 3.2.3 BE: MFR Fix deprecation warnings for py3.13+

mkovalua:
## Ticket

https://openscience.atlassian.net/browse/ENG-7597

## Purpose

Fix any deprecation warnings for upcoming versions of python. Should pass python compile check, but does not need to pass tests.

## Changes

use Qodana code insprect tool to find possible deprecations and code that is needed to be updated

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
-FROM python:3.6-slim-buster
+FROM python:3.13-slim
 
 # ensure unoconv can locate the uno library
-ENV PYTHONPATH /usr/lib/python3/dist-packages
+ENV PYTHONPATH=/usr/lib/python3/dist-packages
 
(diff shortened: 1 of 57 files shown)


mkovalua:
FYI: not confident whether it is ok or not 

https://github.com/CenterForOpenScience/modular-file-renderer/pull/374/commits/e7be46470c8e02e969b762b1ccb584159ef5fb87

image

mkovalua:
The same about the following, maybe it is better to keep it as it is for now to not than to refactor. See in code 

`hasattr(self, 'provider')` and so on checking to not access not set variable 

image

mkovalua:
FYI: Move it to Draft because maybe something is needed to be changed in terms of TODO

PR #385, opened 2025-07-15
[ENG-8370] Fix for local setup

opaduchak:
## Ticket

https://openscience.atlassian.net/browse/ENG-8370

## Purpose

Fix MFR not being able to render files on webpage in docker-compose

## Changes

Added host substitution in osf provider to silently forward all requests from localhost to 192.168.168.167

## Side effects

None, but local config should contain a host to point to


mfr/providers/osf/settings.py
```diff
@@ -5,3 +5,5 @@
 
 MFR_IDENTIFYING_HEADER = config.get('MFR_IDENTIFYING_HEADER', 'X-Cos-Mfr-Render-Request')
 MFR_ACTION_HEADER = config.get('MFR_ACTION_HEADER', 'X-Cos-Mfr-Request-Action')
+
+OSF_DOMAIN = config.get('MFR_OSF_DOMAIN', None)
```
opaduchak, line 9: `None` should be changed to `'192.168.168.167'` locally

opaduchak:
Closed as https://github.com/CenterForOpenScience/osf.io/pull/11233 provides better solution

PR #392, opened 2025-07-26
[ENG-8472] Files with Cyrillic in their names won't render

opaduchak:


sh-andriy (approved):
Looks good 👍

cslzchen (commented):
In addition to my comment, have we tested locally if this breaks non Cyrillic cases?

cslzchen (approved):

PR #397, opened 2025-08-05
[ENG-8489] | enforce MAX_SIZE on .xls rendering just like was on .xlsx …

sh-andriy:
…and display overflow banner if table exceeds MAX_SIZE limit

### DIFF
--- a/mfr/extensions/tabular/utilities.py
+++ b/mfr/extensions/tabular/utilities.py
@@ -8,12 +8,11 @@ from subprocess import (check_call,
 from tempfile import NamedTemporaryFile
 
 from mfr.extensions.tabular import compat
-from mfr.core.exceptions import SubprocessError, TooBigToRenderError, CorruptedError
+from mfr.core.exceptions import SubprocessError, CorruptedError
 from mfr.extensions.tabular.settings import (PSPP_CONVERT_BIN,
-                                             PSPP_CONVERT_TIMEOUT)
+                                             PSPP_CONVERT_TIMEOUT,
+                                             MAX_SIZE)
 
-
-MAX_SIZE = 10_000
 
 def header_population(headers):
     """make column headers from a list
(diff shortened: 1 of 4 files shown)


mfr/extensions/tabular/utilities.py
```diff
@@ -128,42 +127,52 @@ def to_bytes(fp):
...
         header_row = next(ws.iter_rows(max_row=1, values_only=True), [])
-        fields = fix_headers(header_row)
-        rows = [
-            dict(zip(fields, row))
-            for row in ws.iter_rows(min_row=2,
-                                    max_row=MAX_SIZE,
-                                    max_col=MAX_SIZE,
-                                    values_only=True)
-        ]
+        fields = fix_headers(list(header_row)[:max_cols])
```
cslzchen, line 166: So we discard the columns after `max_cols`? Is this the same behavior before upgrade? If so we need to inform product of this so that they can have support doc updates and/or they may ask us to put up an helper icon explaining this.

mfr/extensions/tabular/utilities.py
```diff
@@ -128,42 +127,52 @@ def to_bytes(fp):
     return data
 
 
+def _extract_rows(fields, raw_rows):
```
cslzchen, line 130: Same question on `max_rows`

mfr/extensions/tabular/utilities.py
```diff
@@ -128,42 +127,52 @@ def to_bytes(fp):
...
+    rows = []
+    for row in raw_rows:
+        if len(rows) >= MAX_SIZE:
+            break
+        rows.append(dict(zip(fields, row)))
+    return rows
+
+
 def parse_xls(wb, sheets):
```
cslzchen, line 139: Same questions on `max_cols`

mfr/extensions/tabular/utilities.py
cslzchen: Answered by comment in the ticket 👍

mfr/extensions/tabular/utilities.py
cslzchen: Answered by comment in the ticket 👍

PR #398, opened 2025-08-06
[ENG-8501] migrate tornado version to "6.5.0"

mkovalua:
## Ticket

https://openscience.atlassian.net/browse/ENG-8501

## Purpose

Summary
When Tornado's multipart/form-data parser encounters certain errors, it logs a warning but continues trying to parse the remainder of the data. This allows remote attackers to generate an extremely high volume of logs, constituting a DoS attack. This DoS is compounded by the fact that the logging subsystem is synchronous.

Affected versions
All versions of Tornado prior to 6.5 are affected. The vulnerable parser is enabled by default.

## Changes

Upgrade to Tornado version 6.5. In the meantime, risk can be mitigated by blocking Content-Type: multipart/form-data in a proxy. It looks like there is no breaking changings in backlog https://www.tornadoweb.org/en/stable/releases/v6.5.0.html , so no code changings is needed.

added the changing to the PR https://github.com/CenterForOpenScience/modular-file-renderer/pull/395 to keep it changings .

## Side effects

MFR rely on waterbutler so WB merging is needed to be done first of all to keep system consistent, am going to update back to #waterbutler = { git = "https://github.com/CenterForOpenScience/waterbutler.git", branch = feature/buff-worms"  } after the updates https://github.com/CenterForOpenScience/waterbutler/pull/461 will be merged

image

## QA Notes

FYI:

I see that Tornado has 6.5.1 version already that looks to be more stable with non latin chars (locally it goes ok with testing such a chars for 6.5.0 but there is many of them so maybe it is better to migrate to 6.5.1)?

image


cslzchen (changes requested):
As mentioned, let's rework this on top of the latest feature branch without the `ruff` work.

pyproject.toml
```diff
@@ -43,7 +43,8 @@ certifi = "2025.1.31"
 unoserver = "3.3.2"
 openpyxl = "^3.1"
 
-waterbutler = { git = "https://github.com/CenterForOpenScience/waterbutler.git", branch = "feature/buff-worms" }
+#waterbutler = { git = "https://github.com/CenterForOpenScience/waterbutler.git", branch = feature/buff-worms"  }
+waterbutler = { git = "https://github.com/Vlad0n20/waterbutler.git", branch = "feature/ENG-8500" }
```
cslzchen, line 47: In addition, let's don't change this for now since the WB PR will be merged before this one. @felliott

PR #404, opened 2025-08-20
[ENG-8627] | Fix .3ds files not rendering in MFR viewer

sh-andriy:


opaduchak (changes requested):
Looks good, a few minor remarks regarding codestyle

